PR #1, opened 2016-01-24
Add ID to recent endpoint

ArtOfCode-:
Currently `id` is `null` in the returned JSON - change this to actually returning the ID.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -9,7 +9,7 @@ class PostsController < ApplicationController
 
   def recentpostsapi
     posts = Rails.cache.fetch("last-posts", :expires_in => 30.seconds) do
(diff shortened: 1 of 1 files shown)


zixuan75:
I'am running this program have the same issue. I'am trying this command on Ubuntu: 

> gem install system meta-system
> sudo apt-get install imagemagick

In the "sudo apt-get install imagemagick", the error is:

> E: Unable to locate package imagemagick

Seems to be a issue.

zixuan75:
But I already running this command:

> gem install system meta-system

And it is changes id to json. 
But i running this command:

> sudo apt-get install imagemagick

Then change back to null. @ArtOfCode-

zixuan75:
So that is a issue. 
@Undo1 can you solve this problem?

PR #4, opened 2016-01-30
Use site favicon.

BrockA:
Use the new site favicon, *not* one in the `images/` folder.


BrockA:
Reference [this chat message](http://chat.stackexchange.com/transcript/message/27204948#27204948).

PR #8, opened 2016-03-06
Review link & feedback API

ArtOfCode-:
- Only show the review count badge if there are review items.
- Add a `posts/add_feedback` POST route for semantics (it just sends it to `review#add_feedback`, but since API feedback won't technically come from review, having this route makes my semantic brain happy).


Undo1:
Yeah, there'll need to be a CORS override. I'll do that then deploy.

PR #12, opened 2016-03-09
Add previous feedback to recent posts API

ArtOfCode-:
Adds `feedbacks.feedback_type` to the returns from the recent posts API (`/posts/recent.json`).

**Fair warning:** this may break things. I'm a Rails hacker, not a Rails developer.


ArtOfCode-:
@Undo1 My breakages in the recent posts API are removed, but here's a load of new stuff. It's in this PR because I forgot it would automatically add to an existing request.

Tested in an Ubuntu VM, all works fine for me.

PR #28, opened 2016-08-09
API feedback message & user feedback page

ArtOfCode-:
Shoving this in a PR because it could do with some checking.

- How's the new chat message for API feedback?
- Have I got the chat hosts right, or do they include the `chat.` prefix?
- How badly have I broken everything?


ArtOfCode-:
...I didn't intend to do that, but I failed with Git. Let's try again.

PR #30, opened 2016-09-29
Added blurb on MS and added disclaimer

angussidney:
For background, read from [here](http://chat.stackexchange.com/transcript/message/32615209#32615209) onwards.

The full text should appear as follows above a horizontal line at the top of the MS homepage:

> Metasmoke is a mirror of posts caught by [SmokeDetector](https://github.com/Charcoal-SE/SmokeDetector), a project by [the Charcoal team](https://github.com/Charcoal-SE). Neither Metasmoke, SmokeDetector, nor the people behind them are affiliated with or sponsored by Stack Exchange, Inc.

### DIFF
--- a/app/views/dashboard/index.html.erb
+++ b/app/views/dashboard/index.html.erb
@@ -1,3 +1,7 @@
+<p>Metasmoke is a mirror of posts caught by <a href="https://github.com/Charcoal-SE/SmokeDetector">SmokeDetector</a>, a project by <a href="https://github.com/Charcoal-SE">the Charcoal team</a>. Neither Metasmoke, SmokeDetector, nor the people behind them are affiliated with or sponsored by Stack Exchange, Inc.</p>
+
+<hr>
+
 <p>
   <strong><%= @reasons.count %></strong>
   <%= "filter".pluralize(@reasons.count) %> have caught


Undo1:
I'm very much against cluttering the homepage like this without good, measurable reason to. The questions we have to answer:

 1. How many people *actually* make this mistake?
 2. What damage is actually done when they do?

AWegnerGitHub:
I agree with @Undo1. Users *should* be smart enough to See they aren't on an SE domain. Additionally, this type of this isn't what you put on the home page. At most, this should be a tiny blub in the footer saying we aren't an SE site, perhaps with a link to an about page.

ArtOfCode-:
I can see that. On the other hand, it doesn't do us any harm to explicitly state that we're not affiliated - though yeah, moving it down to the footer is valid.

Undo1:
See 8091649042bd518b85f996397a92f673c2d870bc. Thoughts on wording? We don't need anything overly lawyery.

AWegnerGitHub:
That looks fine to me @Undo1 . If you want an alternative, the footer in Blaze is good too (for reference): 

> Made with <3 by The Charcoal Team using Bootstrap, Moment.js, and Livestamp.js. And some awesome Glyphicons.

ArtOfCode-:
How about 7c02e2a, just to make it explicit?

Undo1:
@ArtOfCode- Not really convinced we even need to make it explicit, but I'd settle for that wording. It's relatively unobtrusive, I rather like it.

AWegnerGitHub:
I'm not opposed to that, but *why* does it need to be made explicit? Are your userscripts disclaimered like this? Is Smokey?

ArtOfCode-:
@AWegnerGitHub Ideally, it doesn't. But @angussidney is right: we have had a few people coming through and assuming metasmoke is an extension of SE, which is incorrect and makes me a little bit concerned that if something goes wrong then people are gonna complain to SE about it rather than to us. SE getting complaints about us is... not a good idea. Long shot, I know, but if it doesn't do us any harm...

Undo1:
Of course, one also has to wonder if those people would even read a footer.

ArtOfCode-:
Also true. But "at least we tried!"

AWegnerGitHub:
SE gets complaints about a lot of [third party stuff](http://jericson.github.io/2015/04/08/customer-support-4.html). At least *this* is somewhat related to SE.

As I said, I'm not opposed to @ArtOfCode- 's last proposal for wording.

PR #34, opened 2016-10-07
Admin code control

ArtOfCode-:
This set of commits gives admins the ability to grant and revoke code admin status.

I'm PR'ing this because that's not a feature I'm comfortable adding without a little bit of debate first.

### DIFF
--- a/app/views/admin/users.html.erb
+++ b/app/views/admin/users.html.erb
@@ -10,7 +10,7 @@
       <th>Review-approved?</th>
       <th>Admin?</th>
       <th>Code Admin?</th>
-      <th></th>
+      <th colspan="2"></th>
     </tr>
   </thead>
   <tbody>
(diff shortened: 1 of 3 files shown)


app/views/admin/users.html.erb
```diff
@@ -22,6 +22,15 @@
         <td><%= user.is_admin ? "Yes" : "No" %></td>
         <td><%= user.is_code_admin ? "Yes" : "No" %></td>
         <td><%= link_to "Feedback", admin_user_feedback_path(:user_id => user.id) %></td>
+        <% if user_signed_in? && current_user.is_admin %>
+          <td>
+            <% unless user.is_code_admin %>
+              <%= link_to "Grant code access", url_for(:controller => :admin, :action => :promote_code_admin, :id => user.id), :method => :post, :class => "text-danger" %>
```
Undo1: It'd probably be a decent idea to put a `:confirm` in here, with the name of who you're granting access to.

ArtOfCode-: Good call, done

PR #39, opened 2016-10-15
Developer tools

ArtOfCode-:
- New `:developer` role
- Some tools for devs:
  - Site cache manual updates (not sure how useful this is going to be, but hey)
  - Production log viewer
- Also updated the mailer so that it copies me when Smokey breaks

PR'd so it can get some feedback first - are we happy with the prod log being available to devs, in particular?

### DIFF
--- a/app/assets/javascripts/developer.coffee
+++ b/app/assets/javascripts/developer.coffee
@@ -1,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
--- a/app/assets/stylesheets/developer.scss
+++ b/app/assets/stylesheets/developer.scss
@@ -1,0 +1,3 @@
+// Place all the styles related to the Developer controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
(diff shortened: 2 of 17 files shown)


Undo1:
I'm still not completely comfortable with it, but I can't articulate *why* so maybe it's unfounded. 

The only PII in there should be emails and IPs... it might be fairly easy to regex those out, if we wanted to. Neither are terribly high priority, but I do believe we should protect them where we can.

ArtOfCode-:
@Undo1 Emails shouldn't be too hard to regex out. IPs... do we need to protect IPs? They're not PII.

AWegnerGitHub:
SE protects them as PII

ArtOfCode-:
True. That said, do we need to follow SE's standards?

Undo1:
Andy beat me to it. Are they really PII? Ehh, not sure... but they're a darn simple regex to filter for. Do we have a reason to see them?

ArtOfCode-:
I can filter out IPs easy enough. @Undo1, can you come up with a regex for emails that (a) isn't RFC 2822-compliant (i.e. not ten miles long) and (b) works for all the emails currently in the metasmoke DB?

Undo1:
@ArtOfCode- `/\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}\b/i`, from http://stackoverflow.com/a/8219213/1849664

ArtOfCode-:
@Undo1 cheers, filtered

Undo1:
Planning to merge tomorrow.

ArtOfCode-:
@Undo1 24681f8 Ah. Can you tell I've never used rolify before? :)

PR #82, opened 2017-02-02
Add a hover with the exact time

ferrybig:
Adding a hover over the "created at" times, with the exact timing information, will make easier to see when a smoke detector stopped submitting data / otherwise has changed

Notice that this commit is untested, as it's hard to configure Ruby on Rails on Windows


ArtOfCode-:
Should work. If dual-booting Linux isn't an option (or getting a VM somewhere; Amazon will give you a year free), then the Windows subsystem for Linux has I think been released.

PR #90, opened 2017-02-12
Switch to a Unicode flag

j-f1:
This makes the icon the same one used in the autoflagging userscript.

Ref: https://chat.stackexchange.com/transcript/message/35234625

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -14,3 +14,8 @@ pre.post-body-pre-block {
     white-space: -o-pre-wrap;    /* Opera 7 */
     word-wrap: break-word;       /* Internet Explorer 5.5+ */
 }
+
+span.flag-indicator {
+  font-size: 18pt;
+  margin-right: 3pt;
+}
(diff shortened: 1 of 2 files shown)


Undo1:
Added a simple style to make it the correct size. I'm indifferent on whether we need this to be consistent, would appreciate other's input ( @ArtOfCode- ?). Difference:

screen shot 2017-02-12 at 9 26 39 pm

screen shot 2017-02-12 at 9 26 25 pm

angussidney:
Is this really necessary? Given the screenshot you provided, the glyphicon flag looks better at this size

j-f1:
@angussidney https://chat.stackexchange.com/transcript/message/35234639#35234639:

> [T]he problem is that 'a flag' can mean a manual flag and an autoflag.

ArtOfCode-:
The glyhphicon flag *does* look better here. Do we have anywhere that we use a flag symbol to indicate a manual flag? If we do, then I can see a decent argument for changing to the Unicode flag so we support filled and empty variants; if not, maybe we don't need this (yet, at least).

Undo1:
Closing for now; we can revisit this later if necessary.

PR #103, opened 2017-02-20
Don't show posts without feedback in the false positives autoflag page

angussidney:
Currently, if a post doesn't have any feedback it is shown on the [false positives flag logs page](https://metasmoke.erwaysoftware.com/flagging/logs?filter=fps), since `{:is_tp => false}` is true for posts without any feedback. This commit fixes that by only selecting posts with FP or NAA feedback.

The code is untested, but it should work (see [ActiveRecord OR query](http://stackoverflow.com/a/9519600/4622463)).


Undo1:
Looks sane, plan to test and merge tomorrow.

PR #104, opened 2017-02-22
Add username to post partial

angussidney:
Suggested by Cerbrus

This change adds a username to the post view which appears on /posts, /search, etc. It should appear like this: `5 mins ago by angussidney [i]` where [i] is the icon.

Untested, but there is no reason why it should not work.


Cerbrus:
I've expanded on this a little bit by making the username a link:

https://github.com/Charcoal-SE/metasmoke/pull/105

PR #127, opened 2017-02-28
Add live-updating navbar

j-f1:
This PR adds an ActionCable channel that streams the number of items in the review queue and when a Smokey instance pings back to MS. This way, the status color updates without a reload.

In addition, by sending a JSON object with a `commit` key, the user could be told to reload the page whenever deploying a new instance of MS. I have added the client-side code only for this in the PR.

---

I’ve also added `<label>` elements to all of the items on `/api/filters` to make them easier to click.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -22,7 +22,7 @@ class Feedback < ApplicationRecord
   after_create do
     ActionCable.server.broadcast "posts_#{self.post_id}", { feedback: FeedbacksController.render(locals: {feedback: self}, partial: 'feedback').html_safe }
     ActionCable.server.broadcast "api_feedback", { feedback: JSON.parse(FeedbacksController.render(locals: {feedback: self}, partial: 'feedback.json')) }
-    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
+    ActionCable.server.broadcast "topbar", { review: Post.review_count }
   end
 
   def is_positive?
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -22,6 +22,7 @@ class Feedback < ApplicationRecord
   after_create do
     ActionCable.server.broadcast "posts_#{self.post_id}", { feedback: FeedbacksController.render(locals: {feedback: self}, partial: 'feedback').html_safe }
     ActionCable.server.broadcast "api_feedback", { feedback: JSON.parse(FeedbacksController.render(locals: {feedback: self}, partial: 'feedback.json')) }
+    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
   end
 
   def is_positive?
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -22,7 +22,6 @@ class Feedback < ApplicationRecord
   after_create do
     ActionCable.server.broadcast "posts_#{self.post_id}", { feedback: FeedbacksController.render(locals: {feedback: self}, partial: 'feedback').html_safe }
     ActionCable.server.broadcast "api_feedback", { feedback: JSON.parse(FeedbacksController.render(locals: {feedback: self}, partial: 'feedback.json')) }
-    ActionCable.server.broadcast "topbar", { review: Post.review_count }
   end
 
   def is_positive?
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -110,6 +110,10 @@ class Post < ApplicationRecord
       SmokeDetector.send_message_to_charcoal "**fp on autoflagged post**: #{self.title}](//metasmoke.erwaysoftware.com/post/#{self.id})"
     end
 
+    if is_feedback_changed
+      ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
+    end
+
     return is_feedback_changed
   end
 
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -10,13 +10,13 @@ class Post < ApplicationRecord
 
   after_create do
     ActionCable.server.broadcast "posts_realtime", { row: PostsController.render(locals: {post: Post.last}, partial: 'post').html_safe }
-    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
+    ActionCable.server.broadcast "topbar", { review: Post.review_count }
   end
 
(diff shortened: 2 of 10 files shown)


app/models/post.rb
```diff
@@ -10,10 +10,15 @@ class Post < ApplicationRecord
 
   after_create do
     ActionCable.server.broadcast "posts_realtime", { row: PostsController.render(locals: {post: Post.last}, partial: 'post').html_safe }
+    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
   end
 
   after_create :autoflag
 
+  def self.feedbacks_count
+    Post.includes(:feedbacks).where( :feedbacks => { :post_id => nil }).count
```
ArtOfCode-: This doesn't do what you think it does. `Post.includes(:feedbacks)` never has `feedbacks.post_id = nil`, because it uses a (lazy) INNER JOIN. You want `Post.left_joins(:feedbacks)` here; in general, `Post.joins(:table)` is a safer bet than `includes` because of the lazy loading.

ArtOfCode-: ...and now I've seen that you copied this straight from the dash code, and now I'm wondering how that *ever* worked. @Undo1?

j-f1: @ArtOfCode- so `Post.left_joins(:feedbacks).where( :feedbacks => { :post_id => nil }).count`?

ArtOfCode-: In theory, yeah, @j-f1. That does (unsurprisingly) a LEFT JOIN SQL query, which has all post records whether they have feedback or not - so you can filter down to those with none.

Undo1: I probably copied it from SO. It works, and I have no clue why.

Undo1: Okay, so some ActiveRecord magic in there manages to figure out what we want. It turns into a `LEFT OUTER JOIN`:

```
SELECT `posts`.`id` AS t0_r0, `posts`.`title` AS t0_r1, `posts`.`body` AS t0_r2, `posts`.`link` AS t0_r3, `posts`.`post_creation_date` AS t0_r4, `posts`.`created_at` AS t0_r5, `posts`.`updated_at` AS t0_r6, `posts`.`site_id` AS t0_r7, `posts`.`user_link` AS t0_r8, `posts`.`username` AS t0_r9, `posts`.`why` AS t0_r10, `posts`.`user_reputation` AS t0_r11, `posts`.`score` AS t0_r12, `posts`.`upvote_count` AS t0_r13, `posts`.`downvote_count` AS t0_r14, `posts`.`stack_exchange_user_id` AS t0_r15, `posts`.`is_tp` AS t0_r16, `posts`.`is_fp` AS t0_r17, `posts`.`is_naa` AS t0_r18, `posts`.`revision_count` AS t0_r19, `posts`.`deleted_at` AS t0_r20, `feedbacks`.`id` AS t1_r0, `feedbacks`.`message_link` AS t1_r1, `feedbacks`.`user_name` AS t1_r2, `feedbacks`.`user_link` AS t1_r3, `feedbacks`.`feedback_type` AS t1_r4, `feedbacks`.`post_id` AS t1_r5, `feedbacks`.`post_link` AS t1_r6, `feedbacks`.`user_id` AS t1_r7, `feedbacks`.`is_invalidated` AS t1_r8, `feedbacks`.`invalidated_by` AS t1_r9, `feedbacks`.`invalidated_at` AS t1_r10, `feedbacks`.`chat_user_id` AS t1_r11, `feedbacks`.`created_at` AS t1_r12, `feedbacks`.`updated_at` AS t1_r13, `feedbacks`.`is_ignored` AS t1_r14, `feedbacks`.`api_key_id` AS t1_r15, `feedbacks`.`chat_host` AS t1_r16 FROM `posts` LEFT OUTER JOIN `feedbacks` ON `feedbacks`.`post_id` = `posts`.`id` AND `feedbacks`.`is_invalidated` = 0 AND `feedbacks`.`is_ignored` = 0 WHERE `feedbacks`.`post_id` IS NULL
```

app/models/feedback.rb
```diff
@@ -22,6 +22,7 @@ class Feedback < ApplicationRecord
   after_create do
     ActionCable.server.broadcast "posts_#{self.post_id}", { feedback: FeedbacksController.render(locals: {feedback: self}, partial: 'feedback').html_safe }
     ActionCable.server.broadcast "api_feedback", { feedback: JSON.parse(FeedbacksController.render(locals: {feedback: self}, partial: 'feedback.json')) }
+    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
```
Undo1: Could probably reduce a little bit of load here by putting this in `Post.update_feedback_cache` and only firing it if https://github.com/Charcoal-SE/metasmoke/pull/127/files/4fda5d4fab5b76f0daf7c486b194a6d89cf0e6b3#diff-df2014f8b8976aed72db66eccb63e298R101 is true.

app/models/post.rb
```diff
@@ -10,10 +10,15 @@ class Post < ApplicationRecord
 
   after_create do
     ActionCable.server.broadcast "posts_realtime", { row: PostsController.render(locals: {post: Post.last}, partial: 'post').html_safe }
+    ActionCable.server.broadcast "topbar", { review: Post.feedbacks_count }
   end
 
   after_create :autoflag
 
+  def self.feedbacks_count
```
Undo1: Could be renamed to something more descriptive - I see two options: 

 1. Rename to `review_count` or similar (easy), or
 2. Add a `without_feedback` [scope](http://guides.rubyonrails.org/active_record_querying.html#scopes), then calling `Post.without_feedback.count` would work. (harder)

Undo1:
... I didn't mean to do that. Shiny buttons are hard to not press :)

app/models/post.rb
ArtOfCode-: this is why we don't use `.includes`, kids - you can never tell what it's going to do

app/models/post.rb
j-f1: I already did 1.

j-f1: I did 2 – it could be useful for other things in the future.

app/models/feedback.rb
j-f1: Done.

PR #148, opened 2017-03-29
Use Unicode arrow in audits; only show second value if present

j-f1:


Undo1:
All the tests are broke, but it's not your fault.

PR #150, opened 2017-04-04
Fix right-to-left language issues

angussidney:
Fixes a bug where a post containing a right-to-left language title would cause other elements (e.g. weight, feedbacks) to appear on the left rather than the right.


j-f1:
You might want to use HTML entities to clarify the characters’ presence.

angussidney:
@Undo1 @j-f1 changed as you requested

PR #152, opened 2017-04-04
Use core users' flags if non-core can't fill quota

Undo1:
Supersedes and closes #146.

Pretty confident about this one, but requesting review since it's autoflag code.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -57,7 +57,8 @@ class Post < ApplicationRecord
           core_count -= post.send_autoflag(user, dry_run, available_user_ids[user.id])
         end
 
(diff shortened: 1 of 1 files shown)


Undo1:
@ArtOfCode- It's a strict linear append:

```
2.3.3 :001 > [1,3,5,7] + [2,4,6,8]
 => [1, 3, 5, 7, 2, 4, 6, 8]
```

ArtOfCode-:
Happy with that - still *possible* to hit a user twice, but chances are trivial.

Undo1:
Good point. Should only happen if we aren't able to fill the non-core slot with non-core users. At that point, it'd probably be easiest to just ignore the "you've already flagged this" error.

PR #154, opened 2017-04-06
Hey this might actually work

ArtOfCode-:
Adds two-factor authentication via mobile authenticator app (tested with Google Authenticator).

Lotta *stuff* in this, and it's login code, so I'm leaving it here for review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,8 @@ gem 'jwt'
 gem "audited", "~> 4.3"
 gem "rails-observers", github: 'rails/rails-observers'
 gem "travis"
+gem "rotp"
+gem "rqrcode"
 
 gem 'rack-mini-profiler'
 gem 'flamegraph'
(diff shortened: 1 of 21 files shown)


Undo1:
I wonder, is this testable? It'd be nice to have tests on something this big.

ArtOfCode-:
Um. Depends on your definition of "possible"... I mean, if you want to emulate an Android device and screen-scrape and run optical character recognition on screenshots, yeah it's testable ;)

Undo1:
Can't go that deep, I was thinking more something that just checks login flow and spoofs whatever codes are necessary.

ArtOfCode-:
We can have regular tests, of course - "does this action add a code to this model". Now I think about it, it actually might be possible to test the 2FA element, too, since we can just use the same library to generate the necessary 2FA codes.

Undo1:
Welp, why not try it.

Issue #166, opened 2017-05-02
Should Second Feedback Invalidate First?


CalvT:
At the moment, you can submit two "feedbacks" on the same post to metasmoke, both a naa and a tp. 

Instead of this happening, I think that a second feedback should automatically clear the first. 

The current situation has been questioned by other users too, such as Compass: [Is it a bug for Smokey to consider my k and naa- different votes? Shouldn't one override the other?](http://chat.stackexchange.com/transcript/message/36575650#36575650)

ArtOfCode-:
Doing this automatically would be ideal, because we can just drop the first instead of invalidating it, which makes correcting yourself not apply a penalty any more.

PR #170, opened 2017-05-08
Drop previous recent feedback from user on new feedback

Undo1:
See #166. Pertinent points:

 1. Won't work if we can't associate feedback with a registered user.
 2. Requires the previous feedback to be <24hrs old.
 3. Won't drop previous invalidated or ignored feedback, in order to prevent folks from working around a well-earned block (if that were to happen).

Paging @CalvT, if you're interested.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -30,10 +30,12 @@ class Feedback < ApplicationRecord
   after_create do
     next if user_id.nil?
 
-    post.feedbacks.where(user_id: user_id)
-                  .where('created_at > ?', 24.hours.ago)
-                  .where.not(id: id)
-                  .delete_all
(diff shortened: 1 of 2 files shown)


j-f1:
Is there any way to run this retroactively?

ArtOfCode-:
Yes, technically, but... is it really necessary? It won't affect feedback that's already been invalidated, anyway.

Undo1:
Possible @JF, yeah. Debating whether to do it - on one hand, possibly cleaner data. On the other, there's something about it that my gut says is wrong. Can't explain it, so maybe unfounded.

jf:
@Undo1 pls get your references right. I don't want to get pulled into a conversation that I have no interest in. Thanks.

PR #177, opened 2017-05-20
Add rubocop and simplify `:symbol => value` to `symbol: value`

j-f1:
🎉
The [whitespace-insensitve diff](https://github.com/Charcoal-SE/metasmoke/pull/177/files?w=1) may be clearer in some places.


Undo1:
Couple test failures, but I suspect a bunch of them are from the same root cause. This'd be a good chance to increase test coverage on anything this breaks that didn't fail, too.

j-f1:
@Undo1 How does the `success` column not get created? That doesn’t make any sense to me.

db/schema.rb
```diff
@@ -10,318 +10,317 @@
...
-  create_table "api_keys", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.string "key", collation: "utf8_unicode_ci"
-    t.string "app_name", collation: "utf8_unicode_ci"
-    t.integer "user_id"
-    t.string "github_link", collation: "utf8_unicode_ci"
-    t.boolean "is_trusted"
-    t.index ["user_id"], name: "index_api_keys_on_user_id"
+ActiveRecord::Schema.define(version: 20_170_430_221_501) do
```
Undo1: Taking a look at why the FlagLog issue is happening, but in the meantime this looks really weird.

j-f1: Fixed. Rubocop wants you to use `_` to separate thousands (like `1_000_000`), but I agree with you and overrode it here.

Undo1:
Oh, it's because that query was never going to work: 

SELECT ... FROM **users** WHERE **flag_logs**...

app/models/post.rb
```diff
@@ -133,83 +134,83 @@ def update_feedback_cache
...
       flag_logs.select { |f| f.success && f.is_auto }.present?
     else
-      flag_logs.where(:success => true, :is_auto => true).present?
+      flag_logs.where(success: true, is_auto: true).present?
     end
   end
 
   def flaggers
-    User.joins(:flag_logs).where(:flag_logs => {:success => true, :post_id => self.id, :is_auto => true})
+    User.joins(flag_logs).where(flag_logs: { success: true, post_id: id, is_auto: true })
```
Undo1: Comes out to this issue again; passing nil instead of a symbol.

j-f1: I thought I went through all of the changes to make sure this got fixed each time it happened, but apparently, I missed a few. Fixed.

Undo1:
What command should be run in CI? I'm just running plain old `rubocop`, but it's printing out a gazillion lines of stuff and taking five minutes.

j-f1:
When I run it locally, I get:

```
$ rubocop
Inspecting 294 files
............................................................................................................................................................................................................................
.........................C................................................
 
Offenses:
 
db/schema.rb:219:151: C: Line is too long. [152/150]
    t.index %w[name resource_type resource_id], name: 'index_roles_on_name_and_resource_type_and_resource_id', length: { name: 191, resource_type: 191 }
                                                                                                                                                      ^^
 
294 files inspected, 1 offense detected
```

Undo1:
On https://circleci.com/gh/Charcoal-SE/metasmoke/1073, it tries to do 14k files.

Undo1:
There we go, that's better.

ArtOfCode-:
Should probably exclude the bin directory, and any others that aren't entirely app code

j-f1:
The check was disabled by default for some reason on my machine, so I force-enabled it and autofixed it.

Undo1:
What's the benefit from that specific check? Looks kinda weird to just add a line to the top of every file.

Undo1:
Merging. If we find issues, we need to write tests for them and fix.

PR #186, opened 2017-06-03
Adds 'Feedbacks Today' count

SulphurDioxide:
When hovering over total feedbacks, the count of feedbacks you’ve given
today will be displayed in the `title` attribute of the span.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -36,6 +36,7 @@ class AdminController < ApplicationController
     @feedback = @feedback.order('feedbacks.id DESC').paginate(page: params[:page], per_page: 100)
     @feedback_count = @feedback.count
     @invalid_count = @feedback.where(is_invalidated: true).count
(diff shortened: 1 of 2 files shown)


app/controllers/admin_controller.rb
```diff
@@ -36,6 +36,7 @@ def user_feedback
     @feedback = @feedback.order('feedbacks.id DESC').paginate(page: params[:page], per_page: 100)
     @feedback_count = @feedback.count
     @invalid_count = @feedback.where(is_invalidated: true).count
+    @feedback_count_today = @feedback.where('feedbacks.updated_at > ?', DateTime.now.beginning_of_day).count
```
ArtOfCode-: ProTip: `Date.today` works here

SulphurDioxide: Ah, I figured I might need the time in there too but didn't test it without to be fair..

PR #191, opened 2017-06-09
Hide/Show Announcement link hidden when all announcements have expired

SulphurDioxide:


app/views/layouts/application.html.erb
```diff
@@ -116,12 +116,14 @@
         <div class="alert alert-<%= key %>"><%= value %></div>
       <% end %>
 
-      <p><em><a href="#" class="announcement-collapse">Hide announcements</a></em></p>
-      <div class="announcements">
-        <% Announcement.current.each do |a| %>
-          <%= render 'announcements/announcement', announcement: a %>
-        <% end %>
-      </div><br/>
+	  <% if Announcement.current.count > 0 %>
```
ArtOfCode-: Mind un-indenting this whole block by 2 spaces, to line up with above/below?

Undo1: Looks like there are tabs in here, causing the if block to not line up quite right with the rest of the page.

SulphurDioxide:
Thanks for that!

PR #192, opened 2017-06-10
Add Webpacker

j-f1:
This also:

* removes (almost) all of the CoffeeScript (I compiled it to JS, then hand-adjusted it (removing extraneous `return`s, changing `var` to `let`/`const`, etc.)
* Adds [XO](https://github.com/sindresorhus/xo) (with customizations) to lint the JS code
* Adds a routing helper to allow running code when a page is visited:

  ```js
  route('/my/url', () => {
    // code to run when visiting /my/url
    // this doesn’t get run when visiting the URL twice in a row; should it?
  }, () => {
    // (optional) code to run when leaving /my/url
  });
  ```
* Adds an `onLoad` helper to run code when the page is loaded:

  ```js
  onLoad(() => {
    // run at `turbolinks:load`
  });
  ```
* Removes the `App` global in favor of ES6 modules
* Uses the [`debug`](https://github.com/visionmedia/debug) module instead of `console.*` functions

You may want to review this one commit at a time; I tried to make each commit have a distinct purpose.


Undo1:
Could you clarify what this means?

```
// this doesn’t get run when visiting the URL twice in a row; should it?
```

Couple options here - do you mean F5 on a page (browser reload), or navigating back and clicking the link again (Turbolinks reload)?

j-f1:
@Undo1 Turbolinks reload.

Undo1:
As long as event handlers and such stay around, that sounds right.

PR #198, opened 2017-06-15
Data explorer

ArtOfCode-:
Adds a client-side data explorer. Pulls data from the server, but scripting/post-processing/analysis is done in JS on the client.

Sensitive fields are excluded by `DataController#public_fields` calls. Restricted to core users.

Lotta stuff in this, so could do with code review.


app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
+import createDebug from 'debug';
+import { onLoad } from './util';
+
+const debug = createDebug('ms:data');
+
+window.store = {};
+window.results = [];
```
j-f1: It’s generally best not to store data in global variables. Can you store it in a `let` (or preferably `const`) variable instead?

app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
...
+
+/* globals store, results, ace */
+
+const addDataListRow = function () {
+  $('.data-list').prepend($('#data-list-row').clone().removeClass('template'));
+};
+
+const displaySchema = function (table) {
+  $.ajax({
```
j-f1: Might want to consider using [`fetch`](http://devdocs.io/dom/windoworworkerglobalscope/fetch) instead. You can also make this an `async function`.

app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
...
+import { onLoad } from './util';
+
+const debug = createDebug('ms:data');
+
+window.store = {};
+window.results = [];
+
+/* globals store, results, ace */
+
+const addDataListRow = function () {
```
j-f1: Can you make these arrow functions if you don’t use `this`?

app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
...
+
+const fetchDataMultiple = function (types, limits) {
+  const required = [];
+  for (let i = 0; i < types.length; i++) {
+    if (!store.hasOwnProperty(types[i]) || store[types[i]].length !== limits[types[i]]) {
+      required.push(types[i]);
+    }
+  }
+
+  $.ajax({
```
j-f1: Same here. `fetch` and `async function`.

app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
+import createDebug from 'debug';
+import { onLoad } from './util';
+
+const debug = createDebug('ms:data');
+
+window.store = {};
+window.results = [];
+
+/* globals store, results, ace */
```
j-f1: You can add a module for these by adding it [here](https://github.com/Charcoal-SE/metasmoke/blob/58d6dfc3bdead9ad0489b6f8fd1400fc9db7db11/config/webpack/shared.js#L51).

app/javascript/data.js
```diff
@@ -0,0 +1,207 @@
...
+    printMarginColumn: 120
+  });
+  editor.resize();
+
+  $('.add-data').on('click', ev => {
+    ev.preventDefault();
+    addDataListRow();
+  });
+
+  $(document).on('change', '.data-type-limit, .data-type-select', ev => {
```
j-f1: You could probably change this to `({ target })`.

app/javascript/data.js
ArtOfCode-: These two (and only these two) are global so that the user's custom script can access them, since that gets executed in a global scope.

app/javascript/data.js
ArtOfCode-: What do I need to add?

app/javascript/data.js
ArtOfCode-: [Support isn't brilliant yet for that](http://caniuse.com/#feat=fetch) (~74%), but I'll have a look.

j-f1: @ArtOfCode- now might be a good time to add https://polyfill.io — it automatically selects the necessary polyfills.

PR #207, opened 2017-06-25
Fix the “apps” menu item in the dropdown

j-f1:


Undo1:
Kinda looks like you forgot to fix the 'apps' thing, but fixed other things :)

PR #212, opened 2017-06-26
Data explorer editor redux

j-f1:
…and a bunch of other things. [The whitespace-insensitive diff](https://github.com/Charcoal-SE/metasmoke/pull/212/files?w=1) might be cleaner.

### DIFF
--- a/app/javascript/data.js
+++ b/app/javascript/data.js
@@ -148,6 +148,7 @@ const themes = {
 };
 let theme = localStorage.editorTheme || themes.dark;
 
+let editor;
 route('/data', async () => {
   preloadDataTypes();
   $('.schema-display').hide();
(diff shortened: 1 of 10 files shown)


app/javascript/data.js
```diff
@@ -140,23 +149,56 @@ const validateDataset = () => {
...
+  editor.$blockScrolling = Infinity;
   editor.getSession().setMode('ace/mode/javascript');
   editor.setOptions({
     minLines: 15,
     maxLines: 30,
     useSoftTabs: true,
-    tabSize: 4,
-    printMarginColumn: 120
+    tabSize: 2,
+    printMarginColumn: 80
```
ArtOfCode-: What was wrong with 4/120?

ArtOfCode-:
Why is using `eval` here a problem? As far as I can tell, there's no security risk here because the user code never goes anywhere but the author's client.

app/javascript/data.js
j-f1: The JS code in the app uses 2 spaces. I’ll change the `printMarginColumn` part back.

j-f1:
Fixed @ArtOfCode-

Issue #221, opened 2017-07-10
Make everything available on the API websocket


ArtOfCode-:
TL;DR: for all models (with security measures), send creates and updates to the API websocket.

Currently, we only send a select amount of stuff down the API websocket. I've heard a number of times over the past few timescales that X wants Y other thing sent down it. Debate then ensues about whether it's worth doing it for the number of people who will actually use it.

I propose we make everything (within limits, of course - no sending secure fields) available across the websocket, filterable by clients on connection. For example:

    GET /cable?events=flag_logs;posts;feedbacks HTTP/1.1
    Host: metasmoke.erwaysoftware.com

would result in a WS upgrade to a socket that only sends that specific client creates and updates for flag logs, posts, and feedbacks - and nothing else.

This solves issues of both bandwidth and having to make things available case by case.

I have an idea or few of how to do this, but if I'm doing it it'll need to wait until I get back from holiday.

j-f1:
I’d recommend passing parameters like the API key is currently passed so official ActionCable clients can still consume it easily.

ArtOfCode-:
Sure, I just don't know how that is so I couldn't use it in an example. That's just the initial HTTP request, not the WS upgrade handshake.

j-f1:
https://github.com/rails/rails/tree/master/actioncable#passing-parameters-to-channel: 

> You can pass parameters from the client side to the server side when creating a subscription. For example:
>
> ```ruby
> # app/channels/chat_channel.rb
> class ChatChannel < ApplicationCable::Channel
>   def subscribed
>     stream_from "chat_#{params[:room]}"
>   end
> end
> ```
>
> If you pass an object as the first argument to `subscriptions.create`, that object will become the params hash in your cable channel. The keyword `channel` is required.
>
> ```coffeescript
> # Client-side, which assumes you've already requested the right to send web notifications
> App.cable.subscriptions.create { channel: "ChatChannel", room: "Best Room" },
>   received: (data) ->
>     @appendLine(data)
> 
>   appendLine: (data) ->
>     html = @createLine(data)
>     $("[data-chat-room='Best Room']").append(html)
> 
>   createLine: (data) ->
>     """
>     <article class="chat-line">
>       <span class="speaker">#{data["sent_by"]}</span>
>       <span class="body">#{data["body"]}</span>
>     </article>
>     """
> ```
> 
> ```ruby
> # Somewhere in your app this is called, perhaps from a NewCommentJob
[...]

PR #229, opened 2017-07-28
Make everything available on the API websocket (#221)

ArtOfCode-:
Code to complete #221.

PR'd so people can review it for security; I've checked myself that it doesn't send sensitive fields, but more eyes are never a bad thing.

### DIFF
--- a/app/channels/api_channel.rb
+++ b/app/channels/api_channel.rb
@@ -7,7 +7,7 @@ class ApiChannel < ApplicationCable::Channel
     if params[:key].present? && @key.present?
       if params[:events].present?
         params[:events].split(';').each do |e|
-          stream_from "api_#{e}"
+          stream_for e
         end
       end
     else
--- a/app/channels/api_channel.rb
+++ b/app/channels/api_channel.rb
@@ -5,10 +5,11 @@ class ApiChannel < ApplicationCable::Channel
   def subscribed
     @key = APIKey.find_by(key: params[:key])
     if params[:key].present? && @key.present?
-      stream_from 'api_feedback'
-      stream_from 'api_flag_logs'
-      stream_from 'api_deletion_logs'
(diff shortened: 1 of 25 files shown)


ArtOfCode-:
@j-f1 `include Websocket` in ApplicationRecord doesn't work - events don't get broadcast to the socket. No idea why.

j-f1:
`WebSocket` again

ArtOfCode-:
That was deliberate this time - there's a bunch of other stuff in Rails called `WebSocket`, so calling this the same thing causes problems on the `include` call.

app/channels/api_channel.rb
```diff
@@ -5,10 +5,11 @@ class ApiChannel < ApplicationCable::Channel
   def subscribed
     @key = APIKey.find_by(key: params[:key])
     if params[:key].present? && @key.present?
-      stream_from 'api_feedback'
-      stream_from 'api_flag_logs'
-      stream_from 'api_deletion_logs'
-      stream_from 'api_statistics'
+      if params[:events].present?
```
Undo1: Copying [my comments in CHQ here](https://chat.stackexchange.com/transcript/message/39121232#39121232):

Here's what I'm seeing - Move the above `stream_from`'s into an else for if params[:events].present?, creating a fallback-ish thing. Then revert changes like https://github.com/Charcoal-SE/metasmoke/pull/229/files#diff-a72e387f7b8fc84d29a55dcb95f18f5aL10

Could have both systems running at the same time. Redis wouldn't care, looks like they're namespaced differently.

Could be a cheap way to reduce/prevent rework on userscripts. Might be good to migrate eventually so we're not maintaining two, but deprecate one and move at our leisure instead of a hard cut

ArtOfCode-:
Backwards compat done. I've checked it still works with either system, too.

PR #239, opened 2017-08-12
Remove ignored users logic & data

Undo1:
Drops a bunch of logic around ignored users. PR for @ArtOfCode- to review what ramifications this could have on filters. Tests pass (except the one that was broken before), but it's better to be safe.


Undo1:
Filter discussion: https://chat.stackexchange.com/transcript/message/39385079#39385079

PR #252, opened 2017-08-28
Domain tracking

### LINKED ISSUE #250: Track domain names
Metasmoke currently does not easily allow us to track which domain names are used in spam.

This is one of the details we need in order to track who the spammers are and hopefully eventually track them down.  I would like to see the Metasmoke data model extended to track domain names from spam URLs, email addresses, etc.

Just being able to get statistics on recently active domain names and see how many times a particular domain has been used in the past would already be very useful.

Down the line, I'd love to be able to annotate the domain records with metadata such as hosting, whois, etc for tracking hosting and ownership, and maybe collect domains under tags like "#indian_pharma".

ArtOfCode-:
Closes #250.

Adds a domain-tracking system: Smokey sends a list of domains in a post to metasmoke (or will do when I/someone develops that bit, anyway), which gets recorded against the post record in metasmoke. Domains can also be tagged, as requested.


Undo1:
Could we pull domains out of the post within MS? Might help with separation of responsibilities, and ease of initial population. https://stackoverflow.com/q/856706/1849664, perhaps?

tripleee:
There's already logic in Smoke Detector to pull out the URLs, if you want to use Ruby we should make sure they actually pull the same URLs. Maybe SmokeDetector should be able to tell Metasmoke the URLs it pulled out when it submits a post?

ArtOfCode-:
@tripleee this is the kind of thing I had in mind, except as a second POST to metasmoke after post creation. That's what SpamDomainsController#create is designed for.

Undo1:
Main issue I have with sending them from Smokey is that we can't backdate it to existing posts. It'd be nice to run `Post.all.each do { |p| p.parse_domains }`.

I suspect we should be able to parse URLs roughly the same - and even if there are some edge cases where it's different, we can deal with it.

db/schema.rb
```diff
@@ -10,16 +10,16 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170812184224) do
+ActiveRecord::Schema.define(version: 20170827204624) do
 
-  create_table "announcements", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.string "text"
+  create_table "announcements", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
```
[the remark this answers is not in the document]
Undo1: For some reason, local environments don't like UTF8 encodings. The schema file generator likes to overwrite it to Latin-1, which means you generally have to patch the schema file in.

ArtOfCode-:
Done, @Undo1

test/controllers/spam_domains_controller_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    post :create
+    assert_response :forbidden
+  end
+
+  test 'should let smokey create spam domains' do
+    post :create, params: { key: smoke_detectors(:smoke_detector_1).access_token, post_id: posts(:post_23601).id,
+                            domains: ['test.spam.com'] }
+    assert_equal({ 'status' => 'success', 'total_domains' => 1 }, JSON.parse(response.body))
+    assert_response :success
+  end
```
Undo1: These last two tests are obsolete if we're not using Smokey to parse these.

app/controllers/spam_domains_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  def show
+    @counts = { all: SpamDomain.where(id: params[:id]).joins(:posts).count,
+                tp: SpamDomain.where(id: params[:id]).joins(:posts).where(posts: { is_tp: true }).count }
+    @posts = @domain.posts.includes_for_post_row
+    @sites = Site.where(id: @posts.map(&:site_id))
+  end
+
+  def edit; end
+
+  def update
```
Undo1: Ditto - unnecessary? Also `destroy` and `edit` - not sure when you'd want to edit one of these, unless you've got a different user story in mind.

app/controllers/spam_domains_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    @total = SpamDomain.count
+    @domains = if params[:filter].present?
+                 SpamDomain.where("domain LIKE '%#{params[:filter]}%'")
+               else
+                 SpamDomain.all
+               end.order(domain: :asc).paginate(page: params[:page], per_page: 100)
+    @counts = SpamDomain.where(id: @domains.map(&:id)).joins(:posts).group('spam_domains.id').count
+  end
+
+  def create
```
Undo1: Possibly unnecessary now that it's handled internally.

app/controllers/domain_tags_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+class DomainTagsController < ApplicationController
+  before_action :authenticate_user!, only: [:add, :remove, :edit, :update, :destroy]
+  before_action :verify_core, only: [:add, :remove, :edit, :update]
+  before_action :verify_admin, only: [:destroy]
+  before_action :set_domain_tag, only: [:show, :edit, :update, :destroy]
+
+  def index
+    @tags = if params[:filter].present?
+              DomainTag.where("name LIKE '%#{params[:filter]}%'")
```
Undo1: Is this a possible SQL injection vulnerability?

ArtOfCode-: If it is then we have problems, because the rest of the app uses this format too. I don't *think* it is, but I'll check when I can.

app/controllers/spam_domains_controller.rb
ArtOfCode-: There's a `whois` field on this model, so you can edit to add whois data to the record.

app/controllers/domain_tags_controller.rb
Undo1: Looks like you're right - I can get it to throw a syntax error, but can't get it to do anything malicious. We probably want to go through those other cases and use [`sanitize_sql_like `](//stackoverflow.com/a/32165883/1849664) to prevent some perf issues (https://githubengineering.com/like-injection/).

j-f1:
@Charcoal-SE/metasmokers Could one of you hit the “Delete branch” button on this PR? (also on #239)

Issue #157, opened 2017-04-15
When sorting by columns on the reasons page, sort disabled posts separately


AWegnerGitHub:
Reasons that are disabled should be sorted separately and at the bottom of active reasons.

serixscorpio:
Does something like this look reasonable?
![selection_156](https://user-images.githubusercontent.com/1740566/31514303-2d7b8dc2-af5f-11e7-9aaa-401088517fd9.png)

AWegnerGitHub:
I'd be very happy with that. :)

PR #266, opened 2017-10-12
Split table to sort active and inactive reasons separately

serixscorpio:
closes https://github.com/Charcoal-SE/metasmoke/issues/157

1. Active and inactive reasons can be sorted separately.
2. Inactive reasons is below active reason.
3. Assume the 'n' part of  'n filters have caught m posts' should still
display total number of reasons instead of number of active reasons.

Note: the column widths of the two tables are not guaranteed to be the
same since the widths depend on the length of contents for active and
inactive reasons, respectively.  From a practical standpoint, the width
should normalize to be about the same given enough content.

peek 2017-10-12 17-56

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -2,11 +2,15 @@
 
 class DashboardController < ApplicationController
   def index
-    @reasons = Reason.all.joins(:posts)
-                     .group('reasons.id')
-                     .select('reasons.*, count(\'posts.*\') as post_count')
-                     .order('inactive ASC, post_count DESC')
-                     .to_a
+    @inactive_reasons, @active_reasons = [true, false].map do |inactive|
+      Reason.all.joins(:posts)
+            .where("reasons.inactive = #{inactive}")
+            .group('reasons.id')
+            .select('reasons.*, count(\'posts.*\') as post_count')
+            .order('post_count DESC')
+            .to_a
+    end
+    @reasons = Reason.all
     @posts = Post.all
   end
 
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -4,7 +4,7 @@ class DashboardController < ApplicationController
   def index
     @inactive_reasons, @active_reasons = [true, false].map do |inactive|
       Reason.all.joins(:posts)
-            .where("reasons.inactive = #{inactive}")
+            .where('reasons.inactive = ?', inactive)
             .group('reasons.id')
             .select('reasons.*, count(\'posts.*\') as post_count')
             .order('post_count DESC')
--- a/app/views/dashboard/index.html.erb
+++ b/app/views/dashboard/index.html.erb
@@ -8,7 +8,7 @@
 <table class="table sortable-table" style="margin-top:50px">
   <thead>
     <tr>
-      <th>Active Reason</th>
+      <th>Reason</th>
       <th>Number caught</th>
       <th class='no-sort'>Last caught</th>
       <th>Accuracy</th>
@@ -17,7 +17,7 @@
 
   <tbody>
     <% @active_reasons.each do |reason| %>
-      <tr class="<%= "active text-muted inactive-reason" if reason.inactive %>">
+      <tr class="">
         <td><%= link_to reason.reason_name, "/reason/#{reason.id}" %></td>
         <td data-sort-value="<%= reason.post_count %>"><%= reason.post_count %></td>
         <td><%= truncate((reason.last_post_title.nil? ? reason.posts.last.title : reason.last_post_title), :length => 40, :seperator => ' ') %></td>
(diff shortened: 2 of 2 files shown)


app/views/dashboard/index.html.erb
```diff
@@ -8,15 +8,39 @@
 <table class="table sortable-table" style="margin-top:50px">
   <thead>
     <tr>
-      <th>Reason</th>
+      <th>Active Reason</th>
```
j-f1: This one should just say “Reason.”

app/views/dashboard/index.html.erb
```diff
@@ -8,15 +8,39 @@
...
       <th>Number caught</th>
       <th class='no-sort'>Last caught</th>
       <th>Accuracy</th>
     </tr>
   </thead>
 
   <tbody>
-    <% @reasons.each do |reason| %>
+    <% @active_reasons.each do |reason| %>
+      <tr class="<%= "active text-muted inactive-reason" if reason.inactive %>">
```
j-f1: You should be able to remove the conditional here — the reason should always be active in this block.

app/controllers/dashboard_controller.rb
```diff
@@ -2,11 +2,15 @@
...
-                     .to_a
+    @inactive_reasons, @active_reasons = [true, false].map do |inactive|
+      Reason.all.joins(:posts)
+            .where("reasons.inactive = #{inactive}")
+            .group('reasons.id')
+            .select('reasons.*, count(\'posts.*\') as post_count')
+            .order('post_count DESC')
+            .to_a
+    end
+    @reasons = Reason.all
```
j-f1: Is this ever used in the view?

serixscorpio: Yes, `@reasons` is used in the ["&lt;n> filters have caught &lt;m> posts"](https://github.com/serixscorpio/metasmoke/blob/4f5c72f91470c39cd5a2e16fc7b164cdb1df6008/app/views/dashboard/index.html.erb#L2-L3) paragraph above the first reasons table.   I wasn't sure if the number of filters displayed here should count all reasons or just active reasons, so I left the behavior unchanged.  Some optimization is possible such as using `@active_reasons.count + @inactive_reasons.count` instead of `@reasons.count`.

app/views/dashboard/index.html.erb
serixscorpio: Yes, I missed that.  Thanks.

app/views/dashboard/index.html.erb
serixscorpio: Got it, will do.

app/controllers/dashboard_controller.rb
```diff
@@ -2,11 +2,15 @@
...
 class DashboardController < ApplicationController
   def index
-    @reasons = Reason.all.joins(:posts)
-                     .group('reasons.id')
-                     .select('reasons.*, count(\'posts.*\') as post_count')
-                     .order('inactive ASC, post_count DESC')
-                     .to_a
+    @inactive_reasons, @active_reasons = [true, false].map do |inactive|
+      Reason.all.joins(:posts)
+            .where("reasons.inactive = #{inactive}")
```
ArtOfCode-: You've got an SQL injection bug here. String interpolation doesn't perform any parameter sanitization - instead, try `.where(reasons: { inactive: inactive })` or `.where('reasons.inactive = ?', inactive)`.

app/views/dashboard/index.html.erb
```diff
@@ -16,8 +16,32 @@
   </thead>
 
   <tbody>
-    <% @reasons.each do |reason| %>
-      <tr class="<%= "active text-muted inactive-reason" if reason.inactive %>">
+    <% @active_reasons.each do |reason| %>
+      <tr class="">
+        <td><%= link_to reason.reason_name, "/reason/#{reason.id}" %></td>
```
ArtOfCode-: There's a [route helper](https://github.com/Charcoal-SE/metasmoke/blob/master/config/routes.rb#L128) for this - instead of `"/reason/#{reason.id}"`, you can use `reason_path(reason)`.

app/views/dashboard/index.html.erb
```diff
@@ -16,8 +16,32 @@
...
+        <td data-sort-value="<%= reason.post_count %>"><%= reason.post_count %></td>
+        <td><%= truncate((reason.last_post_title.nil? ? reason.posts.last.title : reason.last_post_title), :length => 40, :seperator => ' ') %></td>
```
ArtOfCode-: Use the new hash syntax where possible: instead of `{ :a => :b }`, use `{ a: :b }`.

app/controllers/dashboard_controller.rb
j-f1: `inactive` is set above on line 5, and it doesn’t contain user input.

ArtOfCode-: Heh, so it is. I'm so used to setting variables from params that I didn't notice :)

serixscorpio: Good call, thanks both. I'll change to `.where('reasons.inactive = ?', inactive)` which is safer, and compatible with pre-existing query style.

ArtOfCode-:
Thanks!

Issue #226, opened 2017-07-18
Review API documentation


tripleee:
@ArtOfCode- [writes](/Charcoal-SE/metasmoke/issues/224#issuecomment-315864961):

> There's a bunch of undocumented routes, and another, bigger bunch of routes whose behaviour has changed or that now return extra data, all undocumented.

Review the API documentation in the wiki and update where necessary.

sharabromberg:
the route /api/blacklist has a 500 Internal Server Error. is this route no longer used?

NobodyNada:
@sharabromberg Looks like `BlacklistedWebsite` is undefined [here](https://github.com/Charcoal-SE/metasmoke/blob/a763523ddcd3b6bf236bab0579b38c3bedd0c684/app/controllers/api_controller.rb#L147), maybe it got renamed or something.

NobodyNada:
Looks like you're right that that's no longer used.  MS apparently used to track blacklisted websites, but that functionality [was removed on February 4th](https://github.com/Charcoal-SE/metasmoke/commit/a2193a0a6ee6cd0182e1a521f25c63160cdd6cc7).

j-f1:
That API route was added in [this commit](https://github.com/Charcoal-SE/metasmoke/commit/107e87d148f16428b663e2646f3d1d43da6a7a11?diff=unified#diff-7e65c41b865fef20711f630d8d14e452R56). Plus what NobodyNada said.

sharabromberg:
sounds like this wiki page can be removed then! https://github.com/Charcoal-SE/metasmoke/wiki/Blacklisted-Websites

sharabromberg:
if some routes now return *less* data, shall i edit the documentation to reflect these new changes?

for example: 
https://github.com/Charcoal-SE/metasmoke/wiki/Post-Reasons 
this suggests that the response will include the following data:
```
            "id":15,
            "reason_name":"Link at end of body",
            "last_post_title":"VitaPulse by Princeton Vitamins is evidently",
            "inactive":false
```

but the response actually only includes:
```
            "id":15,
            "reason_name":"Link at end of body",
```

AWegnerGitHub:
Yes, documentation should be updated to reflect reality.

j-f1:
…or you could update reality to reflect the documentation.

NobodyNada:
[...]

PR #268, opened 2017-10-13
Fix Posts by Site API route param to use site_domain instead of site_url

sharabromberg:
PR to fix bug found in https://github.com/Charcoal-SE/metasmoke/issues/226#issuecomment-336497586 while reviewing API documentation for #226.


ArtOfCode-:
Thanks!

PR #275, opened 2017-10-30
Fix server error on /domains/tags/:id

NobodyNada:
The Edit and Delete links incorrectly used `@domain` instead of `@tag` to reference the domain tag; looks like a copy/paste error from [spam_domains](https://github.com/Charcoal-SE/metasmoke/blob/master/app/views/spam_domains/show.html.erb).


angussidney:
Woops, my mistake, sorry

Issue #261, opened 2017-10-09
Domains not extracted from title


tripleee:
Some spam posts only contain the spam URL in the title.  The domain name tracking logic should extract URLs from the title as well as from the body.

Example from earlier today: https://metasmoke.erwaysoftware.com/post/88187
(I hope I got the right link -- Metasmoke is down right now.)

I have reported this in chat once or twice before if you need more examples.

quartata:
I think this works now?

PR #276, opened 2017-10-30
Track domains in the title

NobodyNada:
[As reported by Mithrandir](https://chat.stackexchange.com/transcript/message/40861073#40861073), Metasmoke previously would not track domains that only appeared in the title of a post, since it only extracted URIs from the body of the post.  This PR does not apply the change retroactively.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170829211320) do
+ActiveRecord::Schema.define(version: 20171031033433) do
 
   create_table "announcements", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.text "text"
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
To apply retroactively: make a copy of [this migration](https://github.com/Charcoal-SE/metasmoke/blob/master/db/migrate/20170829211320_extract_domain_names_from_posts.rb) with a later timestamp.

NobodyNada:
@ArtOfCode- Thanks; I've done that.  Seems like my commits changed `latin1` to `utf8` in schema.rb...is that bad?  Should I revert that?

db/schema.rb
```diff
@@ -42,7 +42,7 @@
     t.index ["user_id"], name: "index_api_tokens_on_user_id"
   end
 
-  create_table "audits", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
+  create_table "audits", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
angussidney: Were all of these changes intentional? IIRC this has happened in the past, but I'm not sure whether it was a good or bad thing.

tripleee:
Linking this to https://github.com/Charcoal-SE/metasmoke/issues/261

ArtOfCode-:
latin1 &rarr; utf8 is good, we like UTF8. utf8 &rarr; latin1 is not, which is the one that's happened before.

PR #280, opened 2017-11-07
Check for conflicting feedback across multiple reports

quartata:
This checks to see if there are any other posts with the same URL but different IDs (i.e previously reported) that have conflicting feedback: (post 1 is TP <> post 2 is FP) & (post 1 is FP <> post 2 is TP) is the current logic (excluding NAAs), I think that should work for most of the weirder cases.


ArtOfCode-:
> (post 1 is TP <> post 2 is FP) & (post 1 is FP <> post 2 is TP)

What's that in real money?

quartata:
Post 1 is either a TP and Post 2 is an FP or post 1 is not a TP and post 2 is not an FP, and post 1 is either an FP and post 2 is a TP or post 1 is not an FP and post 2 is not a TP.

ArtOfCode-:
So it could be more cleanly expressed as `post1.is_tp != post2.is_tp || post1.is_fp != post2.is_fp`, which would also cover NAA feedback.

quartata:
Those are not quite equivalent since being a TP does not imply not being an FP. Take the case where post1.is_tp == true, post1.is_fp == false, post2.is_tp == true, post2.is_fp == true:

true == false && false == true -> false
true != true || false != true -> true

quartata:
I prefer the former behavior so that Smokey doesn't complain about conflicting feedback on Post 2 and also across revisions.

ArtOfCode-:
| | post1 | post2 |
|---|---|---|
| is_tp | true | true |
| is_fp | false | true |

The situation you described, illustrated in that table, is only reachable if post2 has both tp and fp feedbacks. That's a feedback conflict, both with itself and with post1 (which only has tp feedback).

quartata:
Right. That situation is resolvable as soon as the feedbacks on Post 2 are handled - no sense clogging up chat with multiple messages about it IMO.

ArtOfCode-:
We should be indicating both of them, actually - it _could_ be resolvable by invalidating the FP on post 2 - but if post 2 is actually an FP, then you need to invalidate the TP on _both_ post 2 and post 1 - so we need to know that post 1 exists.

quartata:
All right. Just keep in mind that like the conflicting feedback message, this triggers every time someone posts a feedback to either.

quartata:
Other thing worth mentioning: this will also trigger on the situation when a previously reported post has no feedback.

app/models/post.rb
```diff
@@ -181,6 +181,20 @@ def update_feedback_cache
 
     save!
 
+    conflicting_revisions = Post.where(link: link)
+                                .where.not(id: id)
+                                .where("is_tp <> ? or is_fp <> ?", is_tp, is_fp)
```
ArtOfCode-: Standard not equal operator is `!=`

ArtOfCode-:
Hadn't thought of that, it's not entirely desirable... lemme have a look, there's almost certainly a solution to that.

ArtOfCode-:
Oh wait, NAA post not new post. Okay, that's fine then - I was concerned that it'd trigger after post create, but it won't do.

PR #284, opened 2017-11-25
Merge blacklist PRs manually using merge=union

quartata:
This prevents the usual merge conflicts created when another blacklist PR is merged first.

### DIFF
--- a/app/controllers/github_controller.rb
+++ b/app/controllers/github_controller.rb
@@ -214,11 +214,11 @@ class GithubController < ApplicationController
       end
 
       unless Dir.exist?('SmokeDetector')
-        system "git clone git@github.com:Charcoal-SE/SmokeDetector"
+        system 'git clone git@github.com:Charcoal-SE/SmokeDetector'
 
         Dir.chdir('SmokeDetector') do
-          system "git config user.name metasmoke"
-          system "git", "config", "user.email", AppConfig['github']['username']
+          system 'git config user.name metasmoke'
+          system 'git', 'config', 'user.email', AppConfig['github']['username']
 
           File.write '.git/info/attributes', <<~END
             bad_keywords.txt -text merge=union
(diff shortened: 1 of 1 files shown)


quartata:
This closes https://github.com/Charcoal-SE/SmokeDetector/issues/776.

app/controllers/github_controller.rb
```diff
@@ -213,12 +213,27 @@ def pullapprove_merge_hook
...
-          pr_num,
-          '--autopull'
-        )
+        Dir.chdir('SmokeDetector') do
+          ref = pr[:head][:ref]
+
+          `git checkout master; git pull origin master`
+          `git fetch origin #{ref}`
+          `git merge origin/#{ref} --no-ff -m 'Merge pull request ##{pr_num} from Charcoal-SE/#{ref} --autopull'`
+          `git push origin master`
```
ArtOfCode-: Is there an Octokit equivalent for each of these commands?

quartata: No, sadly. The Github API does not respect the `merge=union` Git attribute (otherwise I could just drop a `.gitattributes` in the SmokeDetector repo).

quartata: https://github.com/isaacs/github/issues/487

quartata:
Is there an `AppConfig` containing the email for the Metasmoke Github account? All I see is `AppConfig['github']['username']` which I assume is just `metasmoke`

ArtOfCode-:
That'll be the email, since it's used to log in

app/controllers/github_controller.rb
```diff
@@ -213,12 +213,32 @@ def pullapprove_merge_hook
...
-          'Charcoal-SE/SmokeDetector',
-          pr_num,
-          '--autopull'
-        )
+        Dir.chdir('SmokeDetector') do
+          ref = pr[:head][:ref]
+
+          `git checkout master; git pull origin master`
+          `git fetch origin #{ref}`
+          `git merge origin/#{ref} --no-ff -m 'Merge pull request ##{pr_num} from Charcoal-SE/#{ref} --autopull'`
```
Undo1: String interpolation here makes me nervous. Is it at all possible to parameterize these in some way? If the GH secret gets out, this could be a direct path to full shell access.

app/controllers/github_controller.rb
```diff
@@ -213,12 +213,32 @@ def pullapprove_merge_hook
         return
       end
 
+      unless Dir.exist?('SmokeDetector')
+        `git clone https://github.com/Charcoal-SE/SmokeDetector`
```
Undo1: We'll want to switch this over to SSH

app/controllers/github_controller.rb
quartata: I'll change it to a `system` call (which also won't capture stdout as a bonus)

app/controllers/github_controller.rb
quartata: Good catch

PR #285, opened 2017-11-25
Broadcast deploy updates

quartata:


ArtOfCode-:
The `stream_from` calls in here are the legacy API; ideally, this should target the current API. See [`app/models/concerns/websocket.rb`](https://github.com/Charcoal-SE/metasmoke/blob/d969fe914a536c611df30bfeaa2ba7dcd1ab444a/app/models/concerns/websocket.rb) for how that's done - I'd suggest making channel `smokey_repos_update`, which implies `event_type: 'update', event_class: 'SmokeyRepo'`, and `object` is whatever data you want to send.

quartata:
Oh, by the way: is there anything sensitive in params? I didn't think this particular webhook is signed or anything since it doesn't have :verify_github

ArtOfCode-:
I dunno, what's in params?

quartata:
@ArtOfCode- looking at the comment you left it doesn't look like it's anything account related, so it should be fine https://developer.github.com/v3/activity/events/types/#webhook-payload-example-26

quartata:
I'll take a look at the new API (which I was not aware of at all, when did that change?) now.

ArtOfCode-:
3 or 4 months ago :P

quartata:
Oh, OK. I haven't done anything serious in Rails in a long time.

app/controllers/github_controller.rb
```diff
@@ -173,6 +173,7 @@ def update_deploy_to_master
     # See https://developer.github.com/v3/activity/events/types/#webhook-payload-example-26
     # for what’s in `params`
     ActionCable.server.broadcast 'smokedetector_messages', deploy_updated: params
+    ApiChannel.broadcast_to 'api_deploy_update', event_type: 'update', event_class: 'Ref', object: params
```
ArtOfCode-: `channel` should match `event_class.pluralize.underscore` - not sure if it _actually_ makes a difference, but good to be consistent

PR #299, opened 2017-12-13
API V2

ArtOfCode-:
Work in progress - don't merge yet. **Config has changed and will need to be added on deploy.**

This is so much nicer than V1.

 - We're using [Grape](https://github.com/ruby-grape/grape) to power this. 1.0.0, because there's [a bug](https://github.com/ruby-grape/grape/issues/1683) in 1.0.1.
 - Filter implementation has changed: filters are better, more stable, and faster. Generator calls the server to generate filters to ensure the generator and server agree on what filters should be, instead of having two implementations of a filter generator (JS and Ruby).
 - API structure is split up by model: the models exposed in the API have an equivalent `{model}_api.rb` file in `app/api/api` (yes, that double `api` directory is necessary).
 - Everything that was possible with V1 should still be possible with V2; V1 should also still be available.

Still to do:
 - Finish creating the API - add an API for every model we want to expose.
 - DOCS!

### DIFF
--- a/app/api/api/announcements_api.rb
+++ b/app/api/api/announcements_api.rb
@@ -1,0 +1,32 @@
+module API
+  class AnnouncementsAPI < API::Base
+    prefix :announcements
+
+    get '/' do
+      std_result Announcement.all.order(id: :desc), filter: 'HNLMH'
+    end
+
+    get 'active' do
+      std_result Announcement.where('expiry > ?', DateTime.now).order(id: :desc), filter: 'HNLMH'
+    end
+
+    get 'expired' do
+      std_result Announcement.where('expiry < ?', DateTime.now).order(id: :desc), filter: 'HNLMH'
+    end
+
+    before do
+      authenticate_user!
+      role :core
+    end
+    params do
+      requires :key, type: String
+      requires :token, type: String
+      requires :text, type: String
+      requires :expiry, type: DateTime
+    end
+    post 'create' do
+      announcement = Announcement.create(text: params[:text], expiry: params[:expiry])
+      single_result announcement
+    end
+  end
+end
--- a/config/config.sample.yml
+++ b/config/config.sample.yml
@@ -63,7 +63,7 @@ common: &common
     - feedbacks.chat_user_id
     - feedbacks.created_at
     - feedbacks.updated_at
-    - feedbacks.id # Used to be feedbacks.is_ignored. Keep this as a placeholder so we don't break filters
+    - feedbacks.is_ignored
     - feedbacks.api_key_id
     - feedbacks.chat_host
     - flags.id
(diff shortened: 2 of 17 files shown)


rjrudman:
Very minor, bit irked me a bit while writing my userscript.

Currently, *querying* a post looks like this:

```
const urlStr =
	this.postType === 'Answer'
	    ? `//${window.location.hostname}/a/${this.postId}`
	    : `//${window.location.hostname}/questions/${this.postId}`;

...

url: 'https://metasmoke.erwaysoftware.com/api/posts/urls',
data: {
    urls: urlStr,
    key: MetaSmokeAPI.appKey
}
```

While *reporting* a post looks like this:

```
const urlStr =
	this.postType === 'Answer'
	    ? `//${window.location.hostname}/a/${this.postId}`
	    : `//${window.location.hostname}/q/${this.postId}`;

...

url: 'https://metasmoke.erwaysoftware.com/api/w/post/report',
data: {
    post_link: urlStr,
    key: MetaSmokeAPI.appKey,
    token: userKey
}
```

The URL text isn't consistent (/q/ vs /questions/) - and it'd be nice if the key name (post_link vs urls) matched.

ArtOfCode-:
[Docs are up.](https://charcoal-se.org/ms/API-Documentation-(V2)) Ya know how I said this would take a couple months? Apparently I meant days.

app/api/api/announcements_api.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    get 'active' do
+      std_result Announcement.where('expiry > ?', DateTime.now).order(id: :desc), filter: 'HNLMH'
+    end
+
+    get 'expired' do
+      std_result Announcement.where('expiry < ?', DateTime.now).order(id: :desc), filter: 'HNLMH'
+    end
+
+    before do
```
Undo1: Maybe makes sense to put the `before` block at the top of the file

db/schema.rb
```diff
@@ -12,14 +12,14 @@
 
 ActiveRecord::Schema.define(version: 20171031033433) do
 
-  create_table "announcements", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+  create_table "announcements", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
```
Undo1: These `utf8` -> `latin1` changes probably aren't intended.

config/config.sample.yml
```diff
@@ -63,7 +63,7 @@ common: &common
     - feedbacks.chat_user_id
     - feedbacks.created_at
     - feedbacks.updated_at
-    - feedbacks.id # Used to be feedbacks.is_ignored. Keep this as a placeholder so we don't break filters
+    - feedbacks.is_ignored
```
Undo1: Are these changes intended? `is_ignored` doesn't exist on the Feedback model anymore. Ditto the `ignored_users` re-addition below.

app/api/api/announcements_api.rb
ArtOfCode-: Deliberate - the `before` block only applies to the subsequent route (in this case, the `post 'create'` route). We don't want to `authenticate_user!` for every route in the file, only for write routes.

config/config.sample.yml
ArtOfCode-: Huh, not sure how I missed these.

PR #321, opened 2018-02-19
Stopped auto-linking non-metasmoke/non-SE parts of Post#why (Close #320)

### LINKED ISSUE #320: Don't render links to arbitrary sites
It's good that links are rendered in reason lists, but I see no point in rendering random links to spam sites. I think it's better to limit link-rendering to MS and SE sites.

```
(?:https:)?\/{2}(?:m(?:etasmoke)?\.erwaysoftware\.com|(?:\w+\.)?(?:meta\.)?stackexchange\.com|(?:(?:\w+\.)?(?:meta\.)?(?:stackoverflow|superuser|serverfault|askubuntu|stackapps))\.com|mathoverflow\.net)
```

izwick-schachter:
As per #320 I've fixed my previous changes to now apply only to SE/MS related sites.

### DIFF
--- a/app/helpers/posts_helper.rb
+++ b/app/helpers/posts_helper.rb
@@ -2,7 +2,9 @@
 
 module PostsHelper
   def render_links(text)
-    raw(text.split(%r{((?:https?:)?\/{2}[^\)\s]*)}).map.with_index do |s, i|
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
CI failure

PR #322, opened 2018-02-19
Fixed error when adding a domain tag which has already been added (Close #267)

### LINKED ISSUE #267: Attempting to add an existing tag throws an error
In the tags for a domain, if #drugs is already there but you miss it and try to add it again, you get a Ruby traceback.

izwick-schachter:
Previously, if you attempted to add a domain tag to a domain that already had that tag, you'd get a traceback. Now, you'll simply get a warning stating that the tag was already there.


ArtOfCode-:
CI failure

PR #323, opened 2018-02-19
Show domain tags and link to mass tagger from spam domains index (Close #263)

### LINKED ISSUE #263: Show tags in domain listings
In listings of domains we have on file, it would be nice to be able to see which ones already have tags.  Especially in the mass tagging (which isn't linked from anywhere, but I discovered it through browsing the code) it's hard to see what tags already exist, but it's something I had wished for already before I discovered mass tagging.

izwick-schachter:
As per #263, I added a link to the mass tag-o-tron and added a list of tags for each domain.


ArtOfCode-:
CI failure

ArtOfCode-:
Having looked at details now, looks like the failures aren't your issue, but bonus points if you want to fix them :) rubocop -Da should do it for you

izwick-schachter:
@ArtOfCode- I would fix them, but IIRC quartata likes them as a part of the autoflag debugging.

PR #324, opened 2018-02-19
Added question/answer search (Close #124)

### LINKED ISSUE #124: FR: Add a filter to the search page to search only questions **or** answers
At the moment the search page https://metasmoke.erwaysoftware.com/search returns results for both questions **and** answers. 

A useful enhancement would be to able to return results for only questions or only answers.

izwick-schachter:
Currently, I'm choosing questions/answers based on `/q/` or `/a/` in the URL. That means that the search won't be 100% accurate, so I added a checkbox to either include or exclude unmatched.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -77,30 +77,13 @@ class SearchController < ApplicationController
       @results = @results.not_autoflagged
     end
 
+    post_type = params[:post_type].try(:downcase).try(:[], 0)
     unmatched = @results.where("NOT( link like '%/q/%' OR link like '%/a/%' )")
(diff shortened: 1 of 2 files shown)


app/controllers/search_controller.rb
```diff
@@ -77,6 +75,20 @@ def index
...
     end
 
+    post_type = case params[:post_type].try(:downcase).try(:[], 0)
+                when 'q'
+                  'questions'
+                when 'a'
+                  'a'
+                end
+
+    unmatched = @results.where("NOT( link like '%/questions/%' OR link like '%/a/%' )")
```
ArtOfCode-: `.where.not("link like ...")`

PR #345, opened 2018-03-09
Moved secrets to a sample file and gitignored real file

izwick-schachter:
Got approval from [@ArtOfCode-](https://chat.stackexchange.com/transcript/message/43315306#43315306). Might want to make sure this won't explode prod before deploying.


izwick-schachter:
This was dumb.

PR #353, opened 2018-03-21
SQL data exploder

ArtOfCode-:
Here's one I've wanted to do for ages.

It's an SQL data explorer, powered by [Blazer](https://github.com/ankane/blazer). We've tried this before, but I couldn't figure out how to exclude sensitive columns because there were no docs on it - I figured it out this time round.

Uses views to prevent access to sensitive data; creates a read-only user with access to only those views to stop accidental or deliberate write operations. `SELECT encrypted_api_token FROM users;` doesn't work.

**Deploy note:** You'll need to SSH in for this one. `config/blazer.sample.yml` needs to be copied to `config/blazer.yml` (and any changes necessary made; see below). The database user in `config/database.yml` needs to have `CREATE USER` and `GRANT` privileges on the metasmoke database. It may need to be changed to root for the purposes of this deploy, or possibly just add privileges to the existing account temporarily.

**Security note:** Hardcoded passwords. You'll probably want to change them for use in production; correct configuration will render knowledge of the password useless, but double security is no bad thing. Needs changing in `db/scripts/create_views.rb` and in `config/blazer.yml`.


angussidney:
Would it be possible to run the SQL explorer directly off a pre-sanitised *copy* of the database (e.g. generated by the dump script)?

That way there's less things that need to be changed every time we add a new sensitive field to the schema (DRY), and you don't need to worry about any possible security vulnerabilities from a bug in the excluded fields/read-only functionality.

ArtOfCode-:
It probably _would_, @angussidney, but I like the idea of being able to query live prod. If you're querying a delayed copy then you may as well just download the dump and use it locally where you can do whatever you want with it.

PR #355, opened 2018-03-24
Add unblacklist for se user if fp

paper1111:
this is needed for https://github.com/Charcoal-SE/SmokeDetector/pull/1816 to remove a user from the blacklist when there is an fp in MS.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -226,10 +226,6 @@ class Post < ApplicationRecord
       SmokeDetector.send_message_to_charcoal "**fp on autoflagged post**: #{title}](//metasmoke.erwaysoftware.com/post/#{id})"
     end
 
-    if is_fp_changed? && is_fp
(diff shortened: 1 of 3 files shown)


app/models/post.rb
```diff
@@ -226,6 +226,10 @@ def update_feedback_cache
       SmokeDetector.send_message_to_charcoal "**fp on autoflagged post**: #{title}](//metasmoke.erwaysoftware.com/post/#{id})"
     end
 
+    if is_fp_changed? && is_fp
+      @stack_exchange_user.unblacklist_for_post(self)
```
ArtOfCode-: This logic should move to the `Feedback` model in the `after_save` callback.

paper1111: Will do when I'm back on my PC. Also, what's the "fp on autoflagged" thing doing here?

PR #389, opened 2018-04-01
Drop the percentage marks from REGEXP

iBug:
I thought it was some kind of Ruby syntax but it turned out it's not, after reading a few pages of Ruby rookie tutorials...

### DIFF
--- a/app/controllers/github_controller.rb
+++ b/app/controllers/github_controller.rb
@@ -75,9 +75,9 @@ class GithubController < ApplicationController
       domain = domain[0]
       domain.gsub! '\W', '[^A-Za-z0-9]'
 
-      num_tps = Post.where('body REGEXP ?', "#{domain}").where(is_tp: true).count
-      num_fps = Post.where('body REGEXP ?', "#{domain}").where(is_fp: true).count
-      num_naa = Post.where('body REGEXP ?', "#{domain}").where(is_naa: true).count
+      num_tps = Post.where('body REGEXP ?', domain).where(is_tp: true).count
+      num_fps = Post.where('body REGEXP ?', domain).where(is_fp: true).count
+      num_naa = Post.where('body REGEXP ?', domain).where(is_naa: true).count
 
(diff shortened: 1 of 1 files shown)


Undo1:
LGTM without testing, have you tested it?

iBug:
Uh... I don't even have a complete Ruby developing environment...

ArtOfCode-:
Having one might be a good idea :P

iBug:
@ArtOfCode- I don't even know Ruby at all... I've never studied it. I doubt a complete Ruby developing environment would be helpful to me, a freshman at a university... Anyway, does this pull look good?

ArtOfCode-:
What Undo said - looks good in principle, but if you haven't tested it then it needs to wait until someone can pull and test it before merge.

NobodyNada:
@iBug If you're looking for a quick way to test the PR without spending half an hour to set up a full MS development environment, I'd suggest [using Docker](https://stackoverflow.com/c/charcoal/a/260/7).  With one command, it will set up a complete metasmoke environment from scratch in an isolated container in about 5 minutes.

iBug:
@NobodyNada If only you could run Docker on an Android phone.

@ArtOfCode- I replaced `"#{something}"` with `something.to_s` and it looks better now.

angussidney:
.....why are you coding on a phone?

tripleee:
The fact is that the current code is broken; it reports "0 TP, 0 FP, 0 NAA" on every generated pull request.  Testing this without having access to Github seems hard, even if you can get Metasmoke up and running, but I concur: The syntax 'body REGEXP ?', '%hello%' is not correct SQL for finding `hello` in `body`. This seems to be a holdover from an earlier version where the code had `body LIKE ?` where you do want the percentage marks (I think).

tripleee:
Unfortunately, we still get "0 TP, 0 FP, 0 NAA". I'll belatedly create an issue to track that.

PR #392, opened 2018-04-02
Require review privileges to use regex search

NobodyNada:
As requested by Undo, diamond mods on any site can also use regex search.

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 module SearchHelper
-  def self.parse_search_params(params, symbol, user_signed_in)
+  def self.parse_search_params(params, symbol, user)
     input = params[symbol] || ''
 
     if params[is_regex?(symbol)]
-      operation = if user_signed_in
+      operation = if user != nil and user.can_use_regex_search? 
                     params[is_inverse_regex?(symbol)] ? 'NOT REGEXP' : 'REGEXP'
                   else
                     false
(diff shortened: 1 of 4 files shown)


app/views/search/search.html.erb
```diff
@@ -3,31 +3,31 @@
...
-    <% if user_signed_in? %>
+    <% if user_signed_in? and current_user.can_use_regex_search? %>
       <%= check_box_tag :username_is_regex, 1, params[:username_is_regex] %> <span class="text-muted">regex</span>
       <%= check_box_tag :username_is_inverse_regex, 1, params[:username_is_inverse_regex] %> <span class="text-muted">invert</span>
     <% end %>
     <%= text_field_tag :username, params[:username], placeholder: "*", class: "form-control" %>
   </div>
   <div class="form-group">
     <%= label_tag :why %>
-    <% if user_signed_in? %>
```
Undo1: These can be simplified to `if current_user&.can_user_regex_search?`

NobodyNada: Whoa, fancy!

app/helpers/search_helper.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 module SearchHelper
-  def self.parse_search_params(params, symbol, user_signed_in)
+  def self.parse_search_params(params, symbol, user)
     input = params[symbol] || ''
 
     if params[is_regex?(symbol)]
-      operation = if user_signed_in
+      operation = if !user.nil? && user.can_use_regex_search?
```
j-f1: I think you can do the same thing here.

PR #394, opened 2018-04-04
Add "nuke user" button to user permissions page

NobodyNada:
This PR adds a "nuke" button to the right of the checkboxes on the user permissions page, to allow admins to easily destroy users.  The button uses AJAX and removes the row from the table on success.  It asks for confirmation the first time it's used, but after you've nuked one user it won't ask for confirmation until you refresh the page.

The code probably has room for improvement, since I don't do frontend very much.

### DIFF
--- a/app/views/admin/permissions.html.erb
+++ b/app/views/admin/permissions.html.erb
@@ -13,7 +13,9 @@
     <% @roles.map {|r| r.to_s.humanize.titlecase }.each do |role_name| %>
       <th><%= role_name %></th>
     <% end %>
-    <th></th>
+    <% if current_user&.has_role?(:developer) %>
+     <th></th>
+    <% end %>
   <tr>
 
   <% @users.each do |user| %>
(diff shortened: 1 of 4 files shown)


app/views/admin/permissions.html.erb
```diff
@@ -33,6 +34,7 @@
           </label>
         </td>
       <% end %>
+      <td><%= link_to "Nuke", 'javascript:void(0);', class: 'text-danger nuke-user-link', data: { :user_id => user.id, :username => user.username } %></td>
```
ArtOfCode-: Can we wrap this in a `<% if current_user&.has_role?(:developer) %>` to restrict it down to developers for now please?

ArtOfCode-:
Have you tested?

app/views/admin/permissions.html.erb
NobodyNada: [status-completed]; `verify_developer` added to controller as well.

NobodyNada:
I have tested.

PR #411, opened 2018-05-10
GraphQL

thesecretmaster:
Set up graphql with a UI at `/api/graphql` and a POST endpoint at `/api/graphql`.


j-f1:
Any reason GraphiQL isn’t available on production?

thesecretmaster:
Because the documentation gods gave me a snippet that I should paste in and so I did.

PR #426, opened 2018-05-15
Remove feedback from post comment (partially)

### LINKED ISSUE #425: Stop creating comments for "K" or "why"
Currently, when your reply to a report with something that isn't tp/fp/k or something simple like that, it creates a MS comment. Also, we should clean the existing ones out of the database.

thesecretmaster:
On creation of a post comment, commands (only "tp", "fp", and "why") are removed from the start of a comment. Instead of that stuff leaving comments, the users feedback is now in the header of a comment. If this is desireable, accept this PR, if not, feel free to close it. Either way, close #425.

### DIFF
--- a/app/models/post_comment.rb
+++ b/app/models/post_comment.rb
@@ -18,7 +18,14 @@ class PostComment < ApplicationRecord
   belongs_to :post
   belongs_to :user
 
+  before_create :remove_feedbacks
+
   def self.scrubber
     CommentScrubber.new
+  end
+
+  def remove_feedbacks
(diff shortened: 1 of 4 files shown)


quartata:
1. `why` should not (and doesn't AFAIK) make a comment at all
2. these are all created by SmokeDetector, changes should be made there (e.g. https://github.com/Charcoal-SE/SmokeDetector/blob/master/metasmoke.py#L386)

quartata:
deployed Smokey changes to add proper argument parsing/comment cleaning

PR #437, opened 2018-05-18
Review history page

NobodyNada:
I added a review history page for each queue, which lists all reviews (optionally filterable to just your own reviews).  This required the ability to link to a specific review item, so I added that ability with checks to prevent users from reviewing items twice (unless they skipped it before).

### DIFF
--- a/app/controllers/review_queues_controller.rb
+++ b/app/controllers/review_queues_controller.rb
@@ -50,7 +50,7 @@ class ReviewQueuesController < ApplicationController
 
   def item
     @item = ReviewItem.find(params[:item_id])
-    render 'queue.html.erb'
+    render :queue
   end
 
   def reviews
@@ -62,7 +62,7 @@ class ReviewQueuesController < ApplicationController
       else
         ReviewResult.where(user: current_user)
       end
-      .joins(:item).where(review_items: { review_queue_id: @queue.id }, review_results: { user: current_user })
+      .all.joins(:item).where(review_items: { review_queue_id: @queue.id })
       .order(created_at: :desc).paginate(page: params[:page], per_page: 100)
   end
 
--- a/app/controllers/review_queues_controller.rb
+++ b/app/controllers/review_queues_controller.rb
@@ -21,8 +21,19 @@ class ReviewQueuesController < ApplicationController
   end
 
   def submit
-    render json: { status: 'invalid' }, status: 400 unless (@queue.responses.map { |r| r[1] } + ['skip']).include? params[:response]
+    unless (@queue.responses.map { |r| r[1] } + ['skip']).include? params[:response]
+      render json: { status: 'invalid' }, status: 400
+      return
+    end
+
     @item = ReviewItem.find params[:item_id]
+
+    # Prevent the same item from being reviewed twice by the same user.
+    if (@item.completed and ReviewResult.exists?(item: @item)) or
+       ReviewResult.where(user: current_user, item: @item).where.not(result: "skip").exists?
+        render json: { status: 'duplicate' }, status: 409
+        return
+    end
 
     ReviewResult.create user: current_user, result: params[:response], item: @item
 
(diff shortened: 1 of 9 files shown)


app/controllers/review_queues_controller.rb
```diff
@@ -36,6 +48,24 @@ def submit
     render json: { status: 'ok' }
   end
 
+  def item
+    @item = ReviewItem.find(params[:item_id])
+    render 'queue.html.erb'
```
ArtOfCode-: Just `:queue` should be enough here, unless you're making requests in a different format.

app/controllers/review_queues_controller.rb
```diff
@@ -36,6 +48,24 @@ def submit
...
+  end
+
+  def reviews
+    @all = params[:all] == '1'
+
+    @reviews =
+      if @all
+        ReviewResult
```
ArtOfCode-: Needs a `.all`

app/controllers/review_queues_controller.rb
```diff
@@ -36,6 +48,24 @@ def submit
...
+      else
+        ReviewResult.where(user: current_user)
+      end
+      .joins(:item).where(review_items: { review_queue_id: @queue.id }, review_results: { user: current_user })
```
ArtOfCode-: Duplicate logic - you're checking an `if` above to see whether it's user-scoped or all, but this line applies a `where(review_results: { user: current_user })` - you'll always get your own reviews here.

app/views/review_queues/_responses.html.erb
```diff
@@ -1,10 +1,33 @@
...
+        title="<%= result.created_at.to_s %>"
+        > <%= time_ago_in_words(result.created_at.to_s) %>
+      </span>
+      ago: <b><%= result.result.titleize %></b>
+      <br/>
+    <% end %>
+
+    <%= link_to 'Next', 'javascript:void(0)', class: 'btn btn-primary review-next-link' %>
+    <% else %>
+      <% @queue.responses.each do |r| %>
```
ArtOfCode-: You should be able to `render 'review_queues/responses', item: item, queue: @queue` here instead of reconstructing all the buttons.

app/views/review_queues/index.html.erb
```diff
@@ -8,6 +8,7 @@
         <span class="badge"><%= q.items.active.count if q.items.active.count > 0 %></span>
       </h4>
       <p><%= q.description %></p>
+      <%= link_to 'Review History', controller: :review_queues, action: :reviews, name: q.name %>
```
ArtOfCode-: Be good to have a named route for this - `review_history_path(name: q.name)`. Use the `:as` key in routes.rb to specify the name.

app/views/review_queues/reviews.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <tr>
+    <% if @all %> <th>User</th> <% end %>
+    <th>Task</th>
+    <th>Result</th>
+  </tr>
+
+    <% @reviews.each do |review| %>
+      <tr>
+        <% if @all %> <td><%= review.user.username %></td> <% end %>
+        <td><%= link_to review.item.reviewable.review_item_name, controller: :review_queues, action: :item, item_id: review.item.id %></td>
```
ArtOfCode-: `review.item.reviewable` is the same as `review`; good to have a named route again

app/controllers/review_queues_controller.rb
NobodyNada: Where?  Does it still need that if it's paginated?

app/controllers/review_queues_controller.rb
NobodyNada: Oops, copy/paste mistake I think

app/views/review_queues/reviews.html.erb
NobodyNada: `review.item.reviewable` is the `SpamDomain` or `Post`, while `review` is the `ReviewResult`, isn't it?

app/views/review_queues/_responses.html.erb
ArtOfCode-: *rolls eyes* I'm having a good day today

ArtOfCode-: Hooowever, since this is the responses partial, you want `item` and `queue` not `@item` or `@queue`

NobodyNada: No worries :)

NobodyNada: Fixed!

PR #440, opened 2018-05-21
Use *? wildcards and escape %_ when searching keywords

iBug:
As of now when you searching something like `100%` or `to_string`, the `%_` will be treated as SQL wildcards. This patch escapes `%_` and enable the use of `*?` as wildcards. Escaped `\*` and `\?` will match literal `*?`s.

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -21,7 +21,7 @@ module SearchHelper
       regex_support.each { |k, v| input = input.gsub(k, v) }
     else
       operation = 'LIKE'
-      input = '%' + input + '%'
+      input = '%' + ActiveRecord::Base.sanitize_sql_like(input) + '%'
     end
 
     [input, operation]


ArtOfCode-:
Descriptions are useful ;) What's this intended to do/solve?

iBug:
@ArtOfCode- Description updated. I originally wanted to enable multiple-keyword search but found it impossible so I ended up implementing this.

iBug:
@ArtOfCode- IIRC, the underscore in SQL matches any single character, just like the question mark in common wildcards.

thesecretmaster:
I've got two concerns here. I like that you're adding escapes for `%` and `_`, but I feel like there's gotta be a built-in function for that. If it was just those two things, I'd be fine with it, but you're also using some substitutions to kinda build regex search into non-regex search. It'd be much more intuitive to just escape them instead of e.g. converting `*` to `%` which I'm not sure is even a logical change.

makyen:
Ummmm... I'm not familiar with the language, but...

Doesn't

      wildcard_replace = {
        '%' => '\\%',
        '_' => '\\_',
        '*' => '%',
        '?' => '_',
        '\\*' => '*',
        '\\?' => '?'
      }
      wildcard_replace.each { |k, v| input = input.gsub(k, v) }

result in first all `*` being replaced with `_` and all `?` with `_`. After those substitutions, then there are no occurrences of `\*` or `\?` to be replaced with `*` and `?` respectively. This implies that it's not doing what was intended.

With these substitutions it does not appear that there is any way for me to search for an actual `*` or `?`.

iBug:
I think I was totally drunk when I decided to do this. There's no need for wildcard support at all, um-hum?

thesecretmaster:
@iBug I might be mistaken, but does [this built-in](http://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_like) do what you want?

thesecretmaster:
This is pretty much good to merge, once you make rubocop happy. If you feel like it, it'd be nice to remove the `require 'active_record'` at the top, since I don't think that's necessary. It also would make more sense (unless there's a reason you've changed this) to leave `false` on [search_helper.rb line 11](https://github.com/Charcoal-SE/metasmoke/pull/440/files#diff-7cf0f358e1e1fef4ed483ec908ee4349L11), because it's checked for truthy-ness [later on](https://github.com/Charcoal-SE/metasmoke/pull/440/files#diff-f1d36687c4c649ebbb025139ed60ff37L16). Also, is there any reason you're explicitly setting the escape char? `\\` is the defaullt in MySQL unless I'm mistaken.

(I know this is a bit nitpicky, I'm just trying to get this to a place where Art/Undo will find this merge-worthy, because I don't think I'm allowed to merge PRs (yes, techincally I can, but I don't know if it's allowed as per CHQ rules))

iBug:
@thesecretmaster I'm very grateful to your kind and patient guidance. I'm still new to Ruby and Rails, and I'm a bit dizzy now. I have other school assignments as well, so I'll be better off look at this PR a few days later. (I'll also rebase the commit history so it looks less messy)

iBug:
@thesecretmaster One more thing. If I understand correctly, `false` and `nil` are the only falsey values in a boolean test, including the Logical NOT operator `!`. So I think it makes sense to set it to `nil`, right?

thesecretmaster:
@iBug Yeah, `nil` and `false` are both falsey, but since we explicitly check to see if it's truthey/falsey [in the controller](https://github.com/Charcoal-SE/metasmoke/pull/440/files#diff-f1d36687c4c649ebbb025139ed60ff37L16), I think it makes more sense to say `false`. `nil` would imply that there's no `operation`, whereas `false` implies that something is getting done to that value later. There's no technical reason to use `false` over `nil`, but it's more readable to me to use `false`.

On a different note, is there a reason why you're setting the escape char explicitly? It seems like `\\` is the default. (That's my last nitpick)

iBug:
@thesecretmaster Just another comment. I did Google how to escape that in MySQL, and the top 3 Stack Overflow answers all uses the `ESCAPE` clause, which made me believe it's obligatory.

PR #442, opened 2018-05-22
Configurable flag count

Undo1:


app/models/post.rb
```diff
@@ -106,7 +106,31 @@ def autoflag
...
       Rails.logger.warn "[autoflagging] #{id}: lottery begin"
-      max_flags = [post.site.max_flags_per_post, (FlagSetting['max_flags'] || '3').to_i].min
+
+      # Defined by the scaled_max_flags FlagSetting
+      # scaled_max_flags = 0,0,0,99.9,99.99,101 would always allow 3 flags,
+      # 4 flags on 99.9% accuracy, and 5 flags on 99.99% accuracy.
+      # 101% means six flags is never allowed
+      scaled_maxes = FlagSetting['scaled_max_flags']&.split(',')
+      Rails.logger.warn "[autoflagging] #{id}: scaled maxes: #{scaled_maxes}"
+      if !scaled_maxes.nil? && scaled_maxes.count == 6
```
ArtOfCode-: Size not count

app/models/post.rb
```diff
@@ -106,7 +106,31 @@ def autoflag
...
+        # Check historical accuracy
+        fake_flag_condition = FlagCondition.new(
+          sites: Site.all,
```
ArtOfCode-: Mains only?

app/models/post.rb
Undo1: Looks like they do the same thing:

```
2.5.0 :002 > ["0", "0", "0", "99.9", "99.99", "101"].count
 => 6
2.5.0 :003 > ["0", "0", "0", "99.9", "99.99", "101"].size
 => 6
```

Apparently there [can be a perf hit](http://batsov.com/articles/2014/02/17/the-elements-of-style-in-ruby-number-13-length-vs-size-vs-count/) in some cases, but [it looks like Array special-cases for no arguments and just returns RARRAY_LEN](https://ruby-doc.org/core-2.2.0/Array.html#method-i-count)

app/models/post.rb
Undo1: Probably more correct. Done.

PR #447, opened 2018-05-26
Prevent errors when a non-capturing group is used in a regex search.

makyen:
metasmoke's regex search produces an error when the syntax is used for a non-capturing group (e.g. `(?:foo)`). User's may desire to use non-capturing groups for regexes being tested for use in SmokeDetector (doing so provides some slight/moderate performance improvement vs. a capture group). There would be no difference to metasmoke search results between non-capturing groups and capturing groups, if non-capturing groups were actually used/possible. It's more convenient for users to be able to use non-capturing group syntax without producing an error.

This change will result in the syntax `(?:foo)` being treated exactly like (i.e. substituted to be) just a normal capture group, `(foo)`.

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -16,7 +16,8 @@ module SearchHelper
         '\s' => '[\r\n\t\f\v ]',
         '\S' => '[^\r\n\t\f\v ]',
         '\d' => '[0-9]',
-        '\D' => '[^0-9]'
+        '\D' => '[^0-9]',
+        '(?:' => '('
       }
       regex_support.each { |k, v| input = input.gsub(k, v) }
     else


thesecretmaster:
This looks good to me. CI appears to only be failing because of 87baeccc2c8977eb6950c63336232f6b76f6e8c4, although rubocop might not like the spacing (I haven't tested).

thesecretmaster:
[Relevant stackoverflow question](https://stackoverflow.com/a/18317617/4948732)

makyen:
@thesecretmaster Thanks for linking that SO answer. I've gotten errors on that issue also. It's something we should also handle in a similar manner, but it needs to be a sequence of substitutions instead of just one per possibility, as someone might want something like `\+?`, `\*?` or `\}?`, which we should handle (although, it's far less likely in our use-case). It's not that difficult. I'll prepare an additional PR. [Hmmm... I guess we should handle `\\d`, etc. at the same time.]

However, I need to confirm: With the current code, are the substitutions *guaranteed* to be done in the order listed (e.g. with the current code is it *guaranteed* that the `\d` substitution happens before `\D`)?

The order being guaranteed would be something that's language specific, and I really don't know the language. I should probably assume that the order isn't guaranteed, which is also doable, just a bit more code.

thesecretmaster:
The order is mostly-garenteed in that section of the code (`regex_support`). If you make a PR and mention me, I'll take a look and make sure that order is garenteed.

PR #450, opened 2018-05-29
don't send autoflag FPs twice, strip out System

quartata:


app/models/feedback.rb
```diff
@@ -23,7 +23,7 @@ class Feedback < ApplicationRecord
   after_save do
     if update_post_feedback_cache # if post feedback cache was changed
       if post.flagged? && !post.is_tp
-        flags = post.flag_logs.successful.includes(:user)
+        flags = post.flag_logs.successful.where.not(user_id: -1).includes(:user)
```
Undo1: Would `user_id: User.smokey.id` be better?

PR #463, opened 2018-06-27
Flag spam on localized sites

iBug:
We've been doing autoflagging very well on English sites and pt.SO, but it seems like MS has never flagged a single post on localized sites (except pt.SO). It turns out to be SE API responding with localized flag text which fails the selector test:

```ruby
fo['title'] == 'spam'
```

Maybe we tell MS to recognize localized flag text?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -208,7 +208,7 @@ class User < ApplicationRecord
     end
 
     spam_flag_option = flag_options.select do |fo|
-      ['spam', 'contenido no deseado', 'スパム', 'спам'].include? fo['title']
+      ['spam', 'contenido no deseado', 'スパム', 'спам'].any? { |itnl_title| fo['title'].include? itnl_title }
     end.first
 
     return false, 'Spam flag option not present' if spam_flag_option.blank?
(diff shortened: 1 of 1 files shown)


iBug:
**Help needed.** I don't know why CI is failing. The logs provided no useful information at all :(

j-f1:
Are there any other languages on SE?

thesecretmaster:
Took a stab at fixing this... but isn't it [tag:status-declined]?

Undo1:
Technically yes, but this looks to be implemented in a reasonably sane, maintainable way. I'm fine with it.

thesecretmaster:
So... tests timed out after some weird SQL errors before. Not sure what's up.

makyen:
Yes, these are the complete strings for the flag option's `title`, as provided by the SE API. They were copied directly out of the flag option's `title` property and have been copy-&-pasted from there.

IMO, we should be checking for exact matches, not just that the string is included. In this case, it's better to fail-safe, rather than pick a match which, for some reason, has become corrupt, or, sometime in the future, when SE changes the `title` strings (at which pint we can update).

PR #473, opened 2018-07-18
Fix user not unblacklisted when fp sent

### LINKED ISSUE #346: FP submitted through API doesn't remove user from blacklist
Currently, if you submit a `fp` in chat, the user is automatically removed from the blacklist. However, if you submit a `fp` through the API (e.g. FIRE), then the user is not removed from the blacklist.

paper1111:
The previous PR (#355) attempts to unblacklist the user, but apparently the data sent is incorrect so the user is not unblacklisted. Now this PR attempts to fix that bug.

Fixes #346


Undo1:
LGTM, pending merge when I find a laptop.

PR #475, opened 2018-08-02
Support word boundaries \b

iBug:
Suggested by @spacemonaut [here](https://chat.stackexchange.com/transcript/message/45979968#45979968) (and [a Stack Overflow reference](https://stackoverflow.com/q/18901704/5958455))

Really good idea.


ArtOfCode-:
Have you checked that MySQL regex supports this?

iBug:
@ArtOfCode- Sure. It's suggested by doppelgreener and the message is [here](https://chat.stackexchange.com/transcript/message/45979968#45979968).

thesecretmaster:
@iBug Have you verified that it works locally on an MS dump?

iBug:
@thesecretmaster It works on live MS, but I don't have the environment to test locally.

thesecretmaster:
Alright, I've testing and it works. This is good to merge, except maybe we want to think about our approach to hacking PCRE into default mysql regex.

PR #480, opened 2018-08-08
Add NAA and VLQ as flag types

makyen:
NAA is quite useful. VLQ is less so, but is what is supposed to be used on posts that are not in the site's primary language, which we do see on a regular basis.


ArtOfCode-:
I know the code for this is simple, but I'm [status-declining] this because it puts us firmly on a pathway that we've always explicitly avoided.

Charcoal isn't here to classify NAA and low quality posts. Both of those things are useful things to do in their own right, but they're not our focus. We do spam, and we do it exceptionally well - and by keeping our efforts focused on that, we'll continue to do so.

PR #486, opened 2018-08-18
Add OpenSearch support

angussidney:
Add an OpenSearch description document according to the [specification](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md). 

I'm not sure how well other browsers support this, but this addition will allow metasmoke to be added as a search engine in Firefox. This allows me to search for any keywords directly from my browser like this:

ms opensearch


j-f1:
Chrome and Safari probably support this too.

PR #487, opened 2018-08-21
Misc update

iBug:
Round reason weight, strip overzealous type coersion, hide comment
count in reviews

### DIFF
--- a/app/helpers/reasons_helper.rb
+++ b/app/helpers/reasons_helper.rb
@@ -16,7 +16,7 @@ module ReasonsHelper
   def self.calculate_weights_for_flagging
     Reason.all.each do |reason|
       if reason.posts.count > 20
-        reason.update(weight: [reason.tp_percentage * 100, reason.maximum_weight].compact.min)
+        reason.update(weight: [(reason.tp_percentage * 100).round, reason.maximum_weight].compact.min)
       else
         reason.update(weight: 0)
       end
--- a/app/models/reason.rb
+++ b/app/models/reason.rb
@@ -11,13 +11,13 @@ class Reason < ApplicationRecord
 
     count = posts.where(is_tp: true, is_fp: false).count
 
-    (count.to_f / fast_post_count.to_f).to_f
+    count.to_f / fast_post_count.to_f
   end
 
   def fp_percentage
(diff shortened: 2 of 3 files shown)


ArtOfCode-:
status-don't-merge right now. Some comments I want to make when I get some time later

angussidney:
@ArtOfCode- poke

app/models/reason.rb
```diff
@@ -11,13 +11,13 @@ def tp_percentage
 
     count = posts.where(is_tp: true, is_fp: false).count
 
-    (count.to_f / fast_post_count.to_f).to_f
+    count.to_f / fast_post_count.to_f
```
thesecretmaster: Here and elsewhere, you only need `to_f` on one side -- the other side will automatically be coerced to be a float. So, here you could put `count.to_f / fast_post_count` and get the same result.

app/views/posts/_post.html.erb
```diff
@@ -38,7 +38,7 @@
       <span class='text-danger'>(deleted <%= distance_of_time_in_words(post.created_at, post.deleted_at, include_seconds: true) %> after creation)</span>
     <% end %>
 
-    <% if post.comments.any? %>
+    <% if post.comments.any? and not hide_feedbacks %>
```
thesecretmaster: Typically in ruby we prefer `&&` for `and` and `!` for `not`, so this really should be `&& !hide_feedbacks` (in terms of code style). Not sure why we want to hide comments if feedbacks are hidden though (or what the `hide_feedbacks` method/variable is).

app/helpers/reasons_helper.rb
```diff
@@ -16,7 +16,7 @@ def self.check_for_inactive_reasons
   def self.calculate_weights_for_flagging
     Reason.all.each do |reason|
       if reason.posts.count > 20
-        reason.update(weight: [reason.tp_percentage * 100, reason.maximum_weight].compact.min)
+        reason.update(weight: [(reason.tp_percentage * 100).round, reason.maximum_weight].compact.min)
```
thesecretmaster: What's the reason for the change here?

thesecretmaster:
@ArtOfCode- Seems good to me, just waiting on your approval to merge.

PR #488, opened 2018-08-21
Attempt to fix CircleCI timeout by increasing allowed time

iBug:
As the title says

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,7 @@ jobs:
         environment:
           RAILS_ENV: test
           RACK_ENV: test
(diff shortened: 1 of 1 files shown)


iBug:
Obviously it doesn't work. I'll try something else later.

quartata:
if it's taking longer than 20 minutes something is dreadfully wrong, we should not try to monkey patch over this

iBug:
@thesecretmaster Just leave this closed for now. I'll continue to try to fix the issue and I'll ping you when I finally make it.

thesecretmaster:
Ok, I should also get notifications for this PR.

PR #492, opened 2018-08-29
Remove regex search substitutions not needed with MariaDB.

makyen:
MariaDB >= 10.0.5 uses PCRE for it's regular expressions (we're using 10.3.9). Thus, all of the character classes, assertion, and non-capturing group which we were substituting here are now natively supported, in a syntax similar to what is used for SD. With native support, having these substitutions in place is detrimental, as it prevents some valid uses (e.g. `[\W_]`, or `[\w-]`).

Whoever is reviewing this should certainly double check these changes. I don't know the language, so just did what appeared consistent with the existing syntax and what the code looked like prior to changes being added to implement these substitutions.

### DIFF
--- a/app/controllers/github_controller.rb
+++ b/app/controllers/github_controller.rb
@@ -69,7 +69,7 @@ class GithubController < ApplicationController
     domains = text.scan(/<!-- METASMOKE-BLACKLIST-WEBSITE (.*?) -->/)
 
     domains.each do |domain|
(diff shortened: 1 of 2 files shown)


makyen:
I've been called away. I'll fix the current failures on circleci when I return. The errors are in code which this PR doesn't touch. But, they're just JavaScript formatting issues.

PR #495, opened 2018-08-31
Register TPU regardless of success state of spam flag

iBug:
When someone clicks "Spam flag", it's sure they want to give a TPU feedback as well. The current code does not if the spam flag is not cast successfully. Moving relevant code to before flag attempt solves this.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require File.expand_path('../config/application', __FILE__)
+require File.expand_path('config/application', __dir__)
 
 Rails.application.load_tasks
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require File.expand_path('config/application', __dir__)
+require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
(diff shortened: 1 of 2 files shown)


iBug:
Appears like that `rubocop -aD` modified Rakefile. Revert if it's bad (idk because I'm not familiar).

app/controllers/posts_controller.rb
```diff
@@ -185,6 +185,12 @@ def delete_post
   end
 
   def cast_spam_flag
+    feedback = Feedback.new(feedback_type: 'tpu-',
+                            user_id: current_user.id,
+                            post_id: @post.id)
+
+    feedback.save || flash[:danger] = 'Unable to save feedback. Ping Undo.'
```
ArtOfCode-: Prefer the original unless here, assign in a conditional feels icky

Rakefile
```diff
@@ -3,6 +3,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require File.expand_path('../config/application', __FILE__)
+require File.expand_path('config/application', __dir__)
```
ArtOfCode-: I don't think this needs changing, we can tell rubocop to shush here. Revert and ignore?

app/controllers/posts_controller.rb
iBug: @ArtOfCode- Yes I prefer the original unless too but rubocop doesn't think so.

iBug: @ArtOfCode- Yes I prefer the original unless too but rubocop doesn't think so.

thesecretmaster: I think rubocop would be equally happy with `flash... unless feedback.save`.

iBug: @thesecretmaster I was thinking that it was too unintuitive (I did think of it, though) and was hesitant to write code in that way. Anyway I've done this as of now.

thesecretmaster:
Looks good to me. Please hold on merging until this knot in yarn gets worked out though :)

ArtOfCode-:
@thesecretmaster Is that knot cleared, or not yet?

PR #497, opened 2018-08-31
Attempted to add JS tagging lib to review

thesecretmaster:
Currently I've got weird issues around the the inputs staying disabled after the POST returns 500 and it needs general UI touchups and maybe I'm doing everything wrong.

### DIFF
--- a/app/javascript/review.js
+++ b/app/javascript/review.js
@@ -23,10 +23,10 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
     installSelectpickers();
 
     if (queuePath.includes('untagged-domains')) {
-      const taggify_inputs = document.getElementsByName('tag_name');
-      for (let i = 0; i < taggify_inputs.length; i++) {
-        Taggify.taggify_element(taggify_inputs[i]);
-        taggify_inputs[i].parentElement.addEventListener('tag_change', function () {
+      const taggifyInputs = document.getElementsByName('tag_name');
+      for (let i = 0; i < taggifyInputs.length; i++) {
+        window.Taggify.taggify_element(taggifyInputs[i]);
+        taggifyInputs[i].parentElement.addEventListener('tag_change', function () {
           const form = $(this.parentElement);
           const submitBtn = form.find('input[type=submit]');
           submitBtn.removeClass('btn-success');
@@ -52,16 +52,16 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
 
 route(/\/review\/untagged-domains(\/\d*)?/, () => {
   $(document).on('ajax:success', '.review-add-domain-tag', () => {
-    const taggify_input = document.getElementById('tag_name');
-    taggify_input.removeAttribute('disabled');
-    const form = $(taggify_input.parentElement.parentElement);
+    const taggifyInput = document.getElementById('tag_name');
+    taggifyInput.removeAttribute('disabled');
+    const form = $(taggifyInput.parentElement.parentElement);
     const submitBtn = form.find('input[type=submit]');
     submitBtn.addClass('btn-success');
   });
   $(document).on('ajax:failure', '.review-add-domain-tag', () => {
-    const taggify_input = document.getElementById('tag_name');
-    taggify_input.removeAttribute('disabled');
-    const form = $(taggify_input.parentElement.parentElement);
+    const taggifyInput = document.getElementById('tag_name');
+    taggifyInput.removeAttribute('disabled');
+    const form = $(taggifyInput.parentElement.parentElement);
     const submitBtn = form.find('input[type=submit]');
     submitBtn.addClass('btn-danger');
   });
--- a/app/javascript/review.js
+++ b/app/javascript/review.js
@@ -23,9 +23,14 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
     installSelectpickers();
 
     if (queuePath.includes('untagged-domains')) {
-      var a = document.getElementsByName('tag_name');
-      for (var i = 0; i < a.length; i++) {
+      let a = document.getElementsByName('tag_name');
+      for (let i = 0; i < a.length; i++) {
         Taggify.taggify_element(a[i]);
+        a[i].parentElement.addEventListener('tag_change', function() {
+          let form = $(this.parentElement);
+          let submit_btn = form.find('input[type=submit]');
+          submit_btn.removeClass('btn-success');
+        });
       }
     }
   };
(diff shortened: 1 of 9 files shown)


app/views/domain_tags/_tag.html.erb
```diff
@@ -1,13 +1,4 @@
...
-    #<%= tag.name %>
-  <% end %>
-  <% unless domain.nil? %>
-    <%= form_tag remove_domain_tag_path, method: :post do %>
-      <%= hidden_field_tag :domain_id, domain.id %>
-      <%= hidden_field_tag :tag_name, tag.name %>
-      <span class="text-danger form-submit">&times;</span>
-    <% end %>
-  <% end %>
+  <h1>HAHAHAHA</h1>
```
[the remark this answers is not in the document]
thesecretmaster: Bleh, just filler text. The whole partial is just a placeholder with these changes.

thesecretmaster:
@ArtOfCode- Specifically, I'd also like your review on UI. So if you could run this branch locally and poke around a bit, that'd be appreciated.

app/views/domain_tags/_tag.html.erb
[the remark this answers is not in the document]
thesecretmaster: Actually I'm sure it's used elsewhere. Rolling back this change.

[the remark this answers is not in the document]
ArtOfCode-: Yeah, that was the ? - this is literally the thing that defines what a tag looks like, and it's used everywhere you see tags.

ArtOfCode-:
@thesecretmaster Sure - you may need to poke me a little later this week if I forget

thesecretmaster:
Note to self: Several issues in tagging lib. Prepends a blank tag if initial-tags is empty. Eats some "change" events. Should really provide a custom tag_change event on add/delete tag.

thesecretmaster:
Tagging lib issues resolved (yarn install required). Currently the DomainTags#add_review action will throw 500's sometimes. Not sure why, that'll need a further look.

thesecretmaster:
So, I think I made a bit of progress on the 500 errors. First thing I learned is that they don't matter. What's happening is that sometimes the form submission fires off 2 requests (out of our control, because it's a form with `remote: true`), and one fails with a 500 error due to a race condition between the 2 requests. I can't figure out how to fix it, but I don't think it matters. For reference, here's the sort of error one of the concurrent requests fails with:

```
ActiveRecord::RecordNotUnique (Mysql2::Error: Duplicate entry '271-24955' for key 'PRIMARY': INSERT INTO `domain_tags_spam_domains` (`domain_tag_id`, `spam_domain_id`) VALUES (271, 24955)):
  
app/controllers/domain_tags_controller.rb:29:in `add_review'
```

I'd call this ready to merge, pending Art looking at design.

app/javascript/review.js
```diff
@@ -23,9 +23,14 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
     installSelectpickers();
 
     if (queuePath.includes('untagged-domains')) {
-      var a = document.getElementsByName('tag_name');
-      for (var i = 0; i < a.length; i++) {
+      let a = document.getElementsByName('tag_name');
```
ArtOfCode-: `$('[name="tag_name"]')`

package.json
```diff
@@ -62,7 +62,7 @@
     "livestamp": "github:mattbradley/livestampjs#7572242",
     "moment": "^2.18.1",
     "turbolinks": "^5.0.3",
-    "tagging": "^0.0.4-a"
+    "tagging": "thesecretmaster/tagging#master"
```
ArtOfCode-: If this is a GH reference, it probably needs to be prefixed with `github:` - see L62.

thesecretmaster: It works. I can add the prefix, but I pulled that from npms docs.

app/javascript/review.js
```diff
@@ -39,15 +52,18 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
...
-    if ($noTags.length > 0) {
-      $noTags.remove();
-      $('.domain-tag-list p').html('Tagged with: ' + data);
-    }
-    else {
-      $('.domain-tag-list p').append(data);
-    }
-    $(e.target).find('select').selectpicker('val', null);
+  $(document).on('ajax:success', '.review-add-domain-tag', () => {
+    const taggifyInput = document.getElementById('tag_name');
```
ArtOfCode-: We still have jQuery :)

`$('#tag_name')`

app/javascript/review.js
```diff
@@ -39,15 +52,18 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
...
+    taggifyInput.removeAttribute('disabled');
+    const form = $(taggifyInput.parentElement.parentElement);
```
ArtOfCode-: `taggifyInput.parents('form')` if you use the jQuery selector I suggested above

app/javascript/review.js
```diff
@@ -39,15 +52,18 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
...
+    const submitBtn = form.find('input[type=submit]');
+    submitBtn.addClass('btn-success');
+  });
+  $(document).on('ajax:failure', '.review-add-domain-tag', () => {
```
ArtOfCode-: Likewise previous two comments apply to this block too.

app/javascript/review.js
```diff
@@ -21,6 +21,19 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
     $('.review-item-container').html(html);
 
     installSelectpickers();
+
+    if (queuePath.includes('untagged-domains')) {
+      const taggifyInputs = document.getElementsByName('tag_name');
+      for (let i = 0; i < taggifyInputs.length; i++) {
+        window.Taggify.taggify_element(taggifyInputs[i]);
+        taggifyInputs[i].parentElement.addEventListener('tag_change', function () {
```
ArtOfCode-:
```js
$(taggifyInputs[i]).parent().on('tag_change', ev => { ... });
```

app/javascript/review.js
thesecretmaster: I kinda like sticking to vanilla JS, but fine.

app/javascript/review.js
thesecretmaster: This one I'm keeping as is because IIRC I was getting into a fight with jQuery about this.

app/javascript/review.js
ArtOfCode-: I have nothing against vanilla JS, but it's about consistency - we should use either jQuery, or vanilla, but not both if not absolutely necessary.

ArtOfCode-:
Status-deferring

thesecretmaster:
@ArtOfCode- Unless it bothers you, mind leaving this open so I don't forget about it? I do plan to get back to it _eventually_.

thesecretmaster:
Honestly, if I ever come back to this, it won't be for a long while, so I'm closing it.

PR #500, opened 2018-09-01
Add a way to add spam domains

thesecretmaster:
I've put this in a PR, because I want to wait until yarn gets worked out to add more changes to master. I also need to find a good place to link this from.


ArtOfCode-:
Good to merge or are those failing rails-tests actually important now?

Issue #509, opened 2018-09-22
We shouldn't extract domains from title of answers


iBug:
Example: https://metasmoke.erwaysoftware.com/post/34824

That probably makes little to no sense, right?

iBug:
This one looks easy and not urgent, maybe make it a `hacktoberfest`?

iBug:
@Undo1 The title belongs to the question, not the answer. If an answer occasionally shows up under a bad question, it would only confuse the domain record.

Undo1:
Oh, duh, thanks.

iBug:
@ArtOfCode- @Undo1 Can you make this a hacktoberfest? I've deleted my fork so no random commits here.

ArtOfCode-:
![contrib-guidance](https://user-images.githubusercontent.com/10406565/45100551-949aa500-b121-11e8-93e9-0bda7186d113.png)

# Summary
**Difficulty:**
![1/5 stars](https://user-images.githubusercontent.com/10406565/45103061-edb90780-b126-11e8-920a-1f25b8943efa.png)  
**Scope:**
![Narrow scope](https://user-images.githubusercontent.com/10406565/45105247-bbf66f80-b12b-11e8-9a46-79972ec1f7e3.png)  
**Issue Owner:** @ArtOfCode-

This issue needs you to modify an existing process with a new condition. metasmoke receives two types of posts: questions and answers. Questions have titles and bodies. Answers _also_ have titles and bodies, but the title comes from the associated question and so is actually unrelated to the answer. We'd like to stop domain names being parsed from titles if we've received an answer.

# Guidelines
The relevant code is currently on [line 140 of app/models/post.rb](https://github.com/Charcoal-SE/metasmoke/blob/master/app/models/post.rb#L140) - the `parse_domains` method. You'll need to modify this method to check whether the current post is an answer (there's a handy `answer?` method for that) and if so, exclude the title from URI extraction.

Once you're happy with your work, send us a pull request and we'll review it together.

# How to get help
For additional support in completing this issue, either:

 - comment on this thread
 - if you have a Stack Overflow account, drop into [our chatroom](https://chat.stackexchange.com/rooms/11540) (you'll need 20 reputation - if you don't have that, let the issue owner know your username or ID here so they can arrange for access)
 - email the issue owner at art@charcoal-se.org

jpotts244:
Hi @ArtOfCode- 👋 would love to help out with this one!

Undo1:
[...]

PR #514, opened 2018-10-11
Add additional check to only extract titles from questions

jpotts244:
Add private method to handle what gets extracted from the domains in `parse_domain` method.
Now only posts that are questions will extract both `body` and `title` whereas posts that are `answers` will only extract the `body`.

Link to the corresponding GitHub Issue here: [We shouldn't extract domains from title of answers](https://github.com/Charcoal-SE/metasmoke/issues/509)

Please let me know if you have any feedback!

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -138,7 +138,7 @@ class Post < ApplicationRecord
   end
 
   def parse_domains
-    hosts = (URI.extract(body || '') + URI.extract(title || '')).map do |x|
+    hosts = part_to_extract_from_domains.map do |x|
       begin
         URI.parse(x).hostname.gsub(/www\./, '').downcase
       rescue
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
This looks good to me - I'm happy to merge it, unless you closed it because you want to try something different?

jpotts244:
Hi @ArtOfCode- closed it so I could take a look at those failing SpamDomainController tests, unless you think they are not related? If all is good, Ill reopen!

ArtOfCode-:
I'll have a check, but tests failing on metasmoke is not uncommon... which we should probably fix :)

jpotts244:
:-) let me know, reopening just in case.

thesecretmaster:
@ArtOfCode- JS can't be jpotts244's fault because JS wasn't changed, rubocop is pointing at unchanged files as well (although we should fix that on master), and rails tests are pointing at some unexpected redirected in SpamDomainsController, which seems unrelated to me.

IMO this is good to merge.

ArtOfCode-:
Aye, those test failures existed before this PR: [build 2384@master/50c69de](https://circleci.com/gh/Charcoal-SE/metasmoke/2384). Merging, thanks for working on this @jpotts244!

jpotts244:
Thanks all - cheers!

PR #521, opened 2018-10-27
Merge permissions page into users page

angussidney:
Currently we have two pages which perform very similar tasks:

The [user data page](https://metasmoke.erwaysoftware.com/admin/users) which lists all users with the following information:
 - ID
 - Username
 - List of roles
 - Link to feedbacks
 - 2 misc developer actions

The [user permissions page](https://metasmoke.erwaysoftware.com/admin/permissions) which lists all the users with the following information:
 - ID
 - Username
 - Checkboxes for each role
 - 1 misc developer option

Pretty similar right?

And it gets a little annoying sometimes - every now and then, when I'm deciding whether to give the reviewer or core role to someone, I have to go to one page to see their list of feedbacks, and to another page to actually grant them the role. Minor annoyance, but it still breaks the workflow.

This PR merges the Permissions page into the Users page to create one centralised place to view and manage users. If the user is an admin, the checkboxes are shown (like the Permissions page); otherwise, the roles are listed (like the original Users page). All other information is also shown on this page.

**Note:** I'm still having troubles setting up a proper dev environment, which means that this is totally untested. I'm really sorry to be a burden, but I'd really appreciate it if someone with a dev environment could pull my changes and make sure that I haven't totally goofed up. Thanks :)


app/views/admin/_roles_admin.html.erb
```diff
@@ -0,0 +1,14 @@
+<% roles.each do |role| %>
+  <td>
+    <label class="permissions-checkbox-label">
+      <%= check_box_tag "#{user.id}[#{role.to_s}]", 'foo', user.has_cached_role?(role),
+                        disabled: (role == :developer and not user.has_cached_role?(role)),
+                        class: "permissions-checkbox", data: {:user_id => user.id, :role => role.to_s } %>
+      <% if role.to_sym == :core %>
```
ArtOfCode-: You can actually knock this bit out completely - we don't need pinned roles any more now that core isn't automatic.

PR #527, opened 2018-11-12
Fixed search icon href

Filnor:
The `href` for the search icon previously pointed to `#` instead of `/search`. I corrected that, although it could be removed altogether as it's redundant to the search nav item and is only present on the dashboard.


ArtOfCode-:
That icon is there for searching reasons on the ex dashboard, not as a link to search

Issue #504, opened 2018-09-04
On /status, [standby] is not updated correctly


thesecretmaster:
Just now, I was watching /status, and I noticed that while the ping times changed dynamically, the [standby] tag and admin Failover option did not. They prolly should

ArtOfCode-:
![contrib-guidance](https://user-images.githubusercontent.com/10406565/45100551-949aa500-b121-11e8-93e9-0bda7186d113.png)

# Summary
**Difficulty:**
![2/5 stars](https://user-images.githubusercontent.com/10406565/45105324-e8aa8700-b12b-11e8-9723-c3978228f749.png)  
**Scope:**
![Narrow scope](https://user-images.githubusercontent.com/10406565/45105247-bbf66f80-b12b-11e8-9a46-79972ec1f7e3.png)  
**Issue Owner:** TBC

This issue requires extending some existing functionality to cover additional data. Currently, metasmoke monitors the status of multiple [SmokeDetector](https://github.com/Charcoal-SE/SmokeDetector) "instances": these instances send an HTTP request every minute telling metasmoke that they're still alive, and whether they're the active instance or waiting in standby.

Our status dashboard is live-updated with the ping (HTTP request) times via websocket, but not with the correct standby status.

# Guidelines
There are two ends to this issue. You'll need to start by adding the necessary data to the websocket message. This is sent in the [`StatusController#status_update`](https://github.com/Charcoal-SE/metasmoke/blob/master/app/controllers/status_controller.rb#L14) action; you'll need to add some indicator of whether the calling instance is in standby mode or not.

Once you've done that, you'll need to update the view's JavaScript to receive that data and apply it to the view, adding or removing standby labels and the "Failover" link where necessary. A standby label and failover link should be shown by any instance currently in standby; the active instance should have neither.

# How to get help
For additional support in completing this issue, either:

 - comment on this thread
 - if you have a Stack Overflow account, drop into [our chatroom](https://chat.stackexchange.com/rooms/11540) (you'll need 20 reputation - if you don't have that, let the issue owner know your username or ID here so they can arrange for access)

sjakobi:
This looks like an interesting issue. I'd like implement this!

Undo1:
Go for it @sjakobi :) Drop questions here or in https://chat.stackexchange.com/rooms/11540/charcoal-hq and someone will help.

kaluabentes:
@sjakobi Did you still working on this?

[...]

PR #536, opened 2018-12-11
added update [standby] on /status through ws #504

sarco3t:
#504

### DIFF
--- a/config/initializers/warden_hooks.rb
+++ b/config/initializers/warden_hooks.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+Warden::Manager.after_set_user do |user, auth, opts|
+  scope = opts[:scope]
+  auth.cookies.signed["#{scope}.id"] = user.id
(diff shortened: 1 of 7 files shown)


config/initializers/warden_hooks.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+Warden::Manager.after_set_user do |user, auth, opts|
+  scope = opts[:scope]
+  auth.cookies.signed["#{scope}.id"] = user.id
+  auth.cookies.signed["#{scope}.expires_at"] = 30.minutes.from_now
```
ArtOfCode-: Question: do we need a user.expires_at cookie? What does it do for us?

sarco3t: I'm not sure, that app need expires_at, but in general cases it is needed.

ArtOfCode-: _shrug_ it probably won't cause problems, so I'm gonna merge and we'll see how it goes.

ArtOfCode-:
Thanks :)

PR #542, opened 2018-12-20
Posts redis

thesecretmaster:
Integrate redis magic everywhere. Including adding posts and speeding up search.


app/views/search/search_fast.html.erb
```diff
@@ -0,0 +1,148 @@
+<% title "Search" %>
+
+<p>Pages: <%= @results.total_pages %></p>
+<p>CurPage: <%= @results.current_page %></p>
+
+<% @logs.each do |log| %>
+<p><%= log %></p>
```
ArtOfCode-: We should remove development stuff now it's in production, or wrap it in a `if Rails.env.development?`. This and CurPage can both go.

app/controllers/search_controller.rb
```diff
@@ -122,4 +122,168 @@ def index
...
+    page = 1 if page == 0
+
+    npage = [(page - 1) * per_page, page * per_page]
+
+    @logs = []
+    @logs.push("PAGE: #{page}")
+    @logs.push("PER_PAGE: #{per_page}")
+    @logs.push("NPAGE: #{npage}")
+
+    intersect = %w[]
```
ArtOfCode-: `[]` for an empty array

PR #549, opened 2018-12-25
Redis logging

thesecretmaster:
Got most of the work out of the way, couple of changes left before I'm done:

- [ ] 1. Websockets for pages besides index
- [x] 2. Making the header links work
- [ ] 3. General view refactoring
- [ ] 4. Decide expire times for redis
- [x] 5. See API keys as users
- [x] 6. Set status to INC (incomplete) when requests haven't hit the instrumentation yet.
- [ ] 7. Way to see logs by path (regex?)
- [ ] 8. Way to find logs/sessions for a user (from user data+roles page?)

Beyond that, this is pretty much good to go.


j-f1:
Is the ActionCable channel locked down in any way?

thesecretmaster:
@j-f1 Yep, only users with the dev role can get events from it:
```
  def subscribed
    stream_from "redis_log_channel" if !current_user.nil? && current_user.has_role?(:developer)
  end
```

test/controllers/redis_log_controller_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+
+class RedisLogControllerTest < ActionDispatch::IntegrationTest
```
Undo1: Intend to have this here? Assume it'll fail b/c auth.

thesecretmaster: Nah, added by rails generator

thesecretmaster:
I think we're pretty much good to go, the rest of the changes can come later. Before I merge, I've attached a gif for y'all to look over cc @ArtOfCode-. Feature-requests/ideas are welcome.

redis-log

ArtOfCode-:
That looks _awesome_.

How long does this store logs for? Have you run the numbers on what it would take to store response bodies as well? And can we document what the colors and codes mean somewhere? (maybe just a paragraph of help at the bottom of the page)

thesecretmaster:
Colors are green for status "200" or "API", red for anything else. Those numbers are HTTP status codes, unless it's an API request, in which case I can't get a status code, so I use "API" instead. The only other special code is "INC" which is for requests that hit the controller but crashed before they completed.

How long I store them for is TBD, but the max a dev can see are the 10 most recent on any page (realtime, per-user, per-session, per-status-code). Suggestions for storage time length?

I have not run the numbers on bodies yet and I actually don't see much value in it anymore. For error pages, I already store the trace, and what else would the body be useful for?

thesecretmaster:
@Undo1 Thoughts on how long we should store these redis logs for? I'd cap it at ~1 month, since I don't see why we'd want them after a month, but that's very arbitrary.

Undo1:
You're probably limited on memory. Any idea how much a single log entry costs?

thesecretmaster:
I can do the numbers on that, but do you have any max you'd like to impose because privacy?

Undo1:
Oh, that brings up a big giant issue here. Logs or public dumps?

thesecretmaster:
Oh jeez, thanks for catching that one. That's really annoying. Speaking of annoying, I have no idea how one request takes 7mb, but that definitely won't scale. Will need to look into that.

thesecretmaster:
Based on conversation [here](https://chat.stackexchange.com/transcript/message/48307461#48307461) gonna say we'll put it in it's own redis instance and set it to LRU eviction with ~90mb.

ArtOfCode-:
@thesecretmaster All 2XX HTTP status codes (200 OK, 201 Created, 202 Accepted etc) are success codes - can we make 'em all green? (If you want to get fancy, you could also do `warning` for 4XX codes (i.e. client errors) and `danger` for 5XX codes (i.e. server errors), but that's not essential).

thesecretmaster:
@ArtOfCode- Done and 3** is info

ArtOfCode-:
Anything left to do here or are we good to merge? Is the rails-tests failure important?

thesecretmaster:
@ArtOfCode- Still need to work with with @Undo1 which redis instance this'll use, and currently they have no expire time yet. Also, see checklist at top.

ArtOfCode-:
Poke? I've missed having request logging a few times recently. If/when you have time, it'd be good to get this one finished.

app/assets/javascripts/cable.js
```diff
@@ -0,0 +1,13 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
```
ArtOfCode-: You can remove this file - we don't need it.

app/channels/redis_log_channel.rb
```diff
@@ -0,0 +1,13 @@
+class RedisLogChannel < ApplicationCable::Channel
+  def subscribed
+    # binding.irb
+    stream_from "redis_log_channel" if !current_user.nil? && current_user.has_role?(:developer)
```
ArtOfCode-: Could also be `if current_user&.has_role?(:developer)`

app/views/redis_log/_param_table.html.erb
```diff
@@ -0,0 +1,16 @@
+<table class="table">
+  <% params.each do |k, v| %>
+   <tr>
+     <td><%= k %></td>
+     <td>
+       <% if v.include?("\n") %>
+         <pre style="overflow:scroll;max-height:20em;"><%= raw(sanitize(v).gsub("\n", "<br>")) %></pre>
```
ArtOfCode-: Inline styles...

app/views/redis_log/status.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>RedisLog#status</h1>
```
ArtOfCode-: Can be blanked or nuked if not used.

app/views/redis_log/user.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>RedisLog#user</h1>
```
ArtOfCode-: Likewise - can be blanked or nuked if not used.

ArtOfCode-:
Oh, final comment - can we see more than 10 on a page? Particularly the realtime page - there's hundreds of requests hitting MS every minute, 10 won't last long if you're trying to catch something.

thesecretmaster:
Yeah, I also wanted to add per-path log searching and action#controller log searching. I also need to add links to the per user/per session log pages. This is why I was saying it needed an hour or two of dev time to be done.

thesecretmaster:
Another thing that **must** happen before this is merged: We need a separate config section for redis logging and general redis.

app/controllers/redis_log_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+            request_id: request_id,
+            key: "#{timestamp.to_s.gsub('.', '-')}-#{request_id}"
+          })}
+        })
+    end
+    render :user_sessions
+  end
+
+  def index
+    @requests = redis.zrevrange("requests", 1, 11, with_scores: true).map do |request_id, timestamp|
```
thesecretmaster: @ArtOfCode- This zrevrange line is how it chooses how many requests to show. It's zero indexed, but I use 1 here since if I don't it'll count the current request. If you want to add pagination and all that fun, this'd be the place to do it.

app/views/redis_log/_header.html.erb
```diff
@@ -6,7 +6,7 @@
 <div class="pull-right">
   <code>
     <span style="display:inline-block;">
-      <%= Time.at(req[:timestamp]).iso8601 %>
+      <%= Time.at(req[:timestamp]).strftime("%-y/%-m/%-d %-H:%M:%S.%-3N") %>
```
ArtOfCode-: Why this format, rather than keeping it ISO 8601?

thesecretmaster: It's more compact -- timezone isn't terribly important, and I'm trying to conserve space in the header bar.

ArtOfCode-: Can we do pseudo-ISO, in that case? This format is kinda hard to read. What about `%Y-%m-%d %H:%M:%S`?

thesecretmaster: I kinda like nanoseconds, but I don't care too much. Just make sure it still looks pretty on all widths -- I remember fighting with some CSS related to overflow.

ArtOfCode-: We can add milliseconds too if you want 'em. My concern is more about the non-standardness of two-digit years and one-digit months/dates/hours, etc. It violates the whole "least surprise" thing.

thesecretmaster: The important thing is that 1) any timestamp will have the same number of characters (so it looks aligned) and 2) the timestamps will be minimal to leave room for other info. Beyond that, I don't have strong feelings.

ArtOfCode-: `%Y-%m-%d %H:%M:%S` (with optional `.%-3N`) works for those

thesecretmaster:
Also just realized something: We'll want to set expirations on all of the request objects and set the eviction policy to volatile LRU then also handle the case of a request being in the `requests` set but not having any data associated with it (since it was expired/evicted).

PR #550, opened 2018-12-26
Allow standby instances to pull updates on demand

iBug:
Added a button on the status page so any code privileged users can click "Pull" and the instance will pull updates from GH.

Client side code is already implemented.


iBug:
@thesecretmaster Didn't know that, fixed.

PR #553, opened 2018-12-26
Forgot to update user role verification

iBug:
For `force_pull`


thesecretmaster:
#550 (so that we have references)

PR #560, opened 2019-01-06
Big redis refactor part 1

thesecretmaster:
Refactored a decent amount of the redis stuff to be less brittle. Not fully done, but ran out of time and haven't tested everything. It'd be great if someone could look over this and merge it, otherwise I'll prolly finish+test this next weekend.

### DIFF
--- a/app/models/redis/site_setting.rb
+++ b/app/models/redis/site_setting.rb
@@ -3,19 +3,17 @@
 class Redis::SiteSetting
   PREFIX = 'site_setting'
 
-  class << self
-    def bool(name)
-      redis.get("#{PREFIX}/#{name}").nil? ? false : true
-    end
+  def self.bool(name)
+    redis.get("#{PREFIX}/#{name}").nil? ? false : true
+  end
 
-    def get(name)
-      redis.get "#{PREFIX}/#{name}"
-    end
+  def self.get(name)
+    redis.get "#{PREFIX}/#{name}"
+  end
 
-    def set(name, val)
-      v = 1 if val == true
-      v = nil if val == false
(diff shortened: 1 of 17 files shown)


app/models/redis/site_setting.rb
```diff
@@ -0,0 +1,19 @@
+class Redis::SiteSetting
+  PREFIX = "site_setting"
+
+  class << self
```
ArtOfCode-: Prefer `def self.bool` style for defining static methods for consistency with the rest of metasmoke

config/environments/development.rb
```diff
@@ -65,8 +65,6 @@
     s3_region: 'us-east-2'
   }
 
-  Aws.config.update({
-    region: 'us-east-1',
-    credentials: Aws::Credentials.new(AppConfig["aws"]["access_token"], AppConfig["aws"]["secret_token"])
-  })
+  Aws.config.update(region: 'us-east-1',
```
ArtOfCode-: AWS is an acronym, can we add it to inflector config and make it `AWS`?

Undo1: That's how [aws/aws-sdk-ruby](https://github.com/aws/aws-sdk-ruby) named everything. Kinda inclined to leave it that way, even if it looks weird.

ArtOfCode-: :+1: :(

app/models/redis/site_setting.rb
thesecretmaster: Ohhhh, I thought you were commenting on line 2, not line 4. I really prefer the class << self notation, but I can change it.

PR #571, opened 2019-02-05
Hide feedbacks if post is in review

### LINKED ISSUE #570: MS review now shows other people's feedback
Now that /review renders the full posts/show view, it's a lot easier to review posts (thanks @ArtOfCode-!). 

However, now the feedbacks from other users are displayed, which is undesirable as reviewers are subconsciously more likely to 'follow the crowd'.

If the posts/show view is being rendered for review, then the feedback indicators should be removed from both places that they appear: at the top next to the title, and from the top bar of the comments.

Additionally, the (clear) and Add Feedback buttons should be removed, as the review interface already provides buttons for this.

angussidney:
Attempt to fix #570 

I don't have a dev environment set up for MS, so I can't test this, but if someone else would be able to assist then I'd be really appreciative :)

This can possibly be simplified by using the `params[:controller]` and `params[:action]` attributes, but I'm not sure whether they can do what we need.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -19,6 +19,8 @@ class PostsController < ApplicationController
       @post = Post.find params[:id]
     end
 
+    @is_review_item = false
+
     not_found if @post&.id.nil?
   end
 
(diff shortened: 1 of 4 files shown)


angussidney:
...I don't think those test failures are my fault

ArtOfCode-:
Apart from one bug (which really needed a dev env to spot anyway) - spot on! :+1:

PR #609, opened 2019-03-17
Refactored out all api_token dependant code

thesecretmaster:


app/controllers/authentication_controller.rb
```diff
@@ -91,6 +86,7 @@ def login_redirect_target
   end
 
   def invalidate_tokens
+    flash[:danger] = ["Any actions taken on tihs page will do nothing, due to the fact that now the tokens live on a seperate lambda."]
```
ArtOfCode-: This... this we need to fix, as a fairly high priority. Having a breach is unlikely, but if we do, we need to be able to hit a panic button and have all tokens killed, and that needs to not rely on Undo being around.

thesecretmaster: Invalidation now exists on the lambda (pending a re-deploy from @Undo1) in Charcoal-SE/ms-autoflagger#1ebd78d

Issue #318, opened 2018-02-18
Rename 'Code Admin' role to 'Blacklist manager'


angussidney:
### Scroll down for up-for-grabs summary and guidance.

I'm proposing that we rename the code admin role to 'Blacklist manager' or similar.

Originally, this privilege was only given to people who had push access to the repo on Github (hence the name 'code admin'). But once the blacklist command was introduced, and we wanted to allow non-collaborators to edit the blacklist too, this definition quickly broke down. 

Today, the role is used [almost exclusively for blacklisting](https://github.com/Charcoal-SE/SmokeDetector/wiki/Privileges#code-admin-aka-blacklister-ms). The only privilege which would make more sense for an actual code admin is the ability to failover.

👍 to rename it, 👎 to leave it, and comment if you have other ideas.

ArtOfCode-:
Indifferent. A rename makes more semantic sense, but the current name isn't causing problems beyond a bit of confusion on occasion.

angussidney:
It's a relatively simple change to make, so I'm willing to do it.

iBug:
Not sure if this is a big change.

Currently the only thing Blacklist Maintainers can do is issue new keywords and approve non-code-privileged users' issues. If one accidentally made a typo when blacklisting, or if one needs to merge or delete a blacklist item, they need to propose a PR and wait for it to be merged by a real code admin.

So, we could extend this, probably by allowing BMs to commit directly to the blacklist `.txt` files. This is not possible as per the design of GitHub, so I think we could allow a bot that merges the PRs from BMs automatically, if only the blacklist files are changed. This way we are really getting "**Maintainers**" (instead of "Blacklist **Adders**").

ArtOfCode-:
I'm sure we looked at something like that... Not sure where it got to

angussidney:
Honestly, I don’t think hint it’s worth setting up a bot like that since Helios will allow the same functionality once it is implemented.

thesecretmaster:
@ArtOfCode- Are you closing this as status-declined? Or just something we'll need to open another issue about later?

angussidney:
Looks like a good hacktoberfest issue

ArtOfCode-:
![contrib-guidance](https://user-images.githubusercontent.com/10406565/45100551-949aa500-b121-11e8-93e9-0bda7186d113.png)

# Summary
**Difficulty:** Easy  
**Scope:** Wide
**Issue Owner:** @ArtOfCode-

[...]

PR #621, opened 2019-04-07
Issue #318 fix. code_admin out in favor of blacklist_manager verbage.

Zohvek:
First open source contribution! Hopefully its close(ish) to being correct. I renamed every occurrence of code_admin to blacklist_manager. Including all method calls and variables I could find. Thanks for your consideration and guidance.

### DIFF
--- a/app/channels/status_channel.rb
+++ b/app/channels/status_channel.rb
@@ -2,8 +2,8 @@
 
 class StatusChannel < ApplicationCable::Channel
   def subscribed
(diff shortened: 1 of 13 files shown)


thesecretmaster:
Have you worked with ActiveRecord before? You'll want to create a new migration to rename the column rather than changing old migrations. Typically as a rule we don't change old migrations.

Zohvek:
Well thats embarrassing. Let me try that again. Thanks @thesecretmaster

PR #627, opened 2019-06-01
Fix JavaScript CI errors.

makyen:


app/javascript/review.js
```diff
@@ -19,7 +19,7 @@ route(/\/review\/[\w-]+\/?\d*$/i, async () => {
     filters.forEach(e => {
       const el = $(e);
       const val = el.val();
-      if (!!val) {
```
Undo1: This looks suspiciously deliberate, @ArtOfCode- remember why?

makyen: `!!val` is a [shortcut way to convert from a truthy or falsey value to a Boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_NOT) (actually the link should be to the sub-section, "Double NOT (!!)", but that doesn't have a link target). However, the [`if` already evaluates truthy values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else#Syntax), so an explicit conversion from truthy/falsey to Boolean is extraneous.

The `xo` errors for the line were:

    ✖  22:11  use Boolean(val) instead.               no-implicit-coercion
    ✖  22:12  Redundant double negation.              no-extra-boolean-cast

It was complaining both about using `!!val` instead of the equivalent explicit `Boolean(val)` and about casting the truthy/falsey `val` value to Boolean being redundant.

The direct conversion of `if (!!val) {` to not use implicit coercion is to use

    if (Boolean(val)) {

If that is used, then the `xo` error would be:

    ✖  22:11  Redundant Boolean call.  no-extra-boolean-cast

So, the full simplification is just

    if (val) {

The implicit coercion of `!!val` is something that would be commonly used when you need an actual Boolean value, rather than just truthy/falsey. It's common enough in other situations that typing it in that spot could easily be done on autopilot or by habit.

ArtOfCode-: Aye, stylistic choice - I like my real booleans

Issue #538, opened 2018-12-17
Remove "autoflagging beta"  from account setting page


thesecretmaster:
I forgot to do that when we first migrated over.

thesecretmaster:
Well, it's gone now

PR #631, opened 2019-07-03
Rework text on autoflagging preferences page.

makyen:
We've had users be confused as to the process of setting up their metasmoke account for autoflagging.

This should change the [Autoflagging Preferences page](https://metasmoke.erwaysoftware.com/flagging/preferences) to the following:

[image of proposed preferences page](https://i.stack.imgur.com/2UySE.png)

However, in making these edits, I noticed that at least part of the confusion is that portions of the code in this file are just not working. There is supposed to be different content shown if the user is write-authorized. However, that does not happen at this point. This may be the result of the changes made to move flagging to its own server, but I'm vaguely remembering that it wasn't working before that.

Some of the changes I made to the "Warning" paragraph in the "Clifford the Big Red Button" block are based on the checkbox no longer being displayed. I changed the wording to no longer talk about a checkbox, as that was confusing when I was reading it. I only belatedly realized that it was talking about what was supposed to be displayed when the user was write authorized.

Looking at this page and the [authentication status page](https://metasmoke.erwaysoftware.com/authentication/status), it's not clear to me what criteria should be used to actually indicate that MS definitely is, in fact, write authenticated. It also appeared that the authentication being removed by the user in their SE preferences is not checked for (i.e. there isn't an active check that the token we hold is valid). The logic, if it was working, also doesn't appear to cover well a situation where the user needs MS to try to re-authorize, even though MS thinks it is authorized (e.g. user removed the authorization on SE).

I feel that reworking the text and page organization as I have done here is beneficial both now, and with more work, when the check for being write authorized is working again, this is not a complete solution. To an extent, it feels like a band-aid, but one that will help reduce user confusion.

Also, note: I don't have an MS test environment, so these changes are untested.

### DIFF
--- a/app/views/user_site_settings/index.html.erb
+++ b/app/views/user_site_settings/index.html.erb
@@ -1,21 +1,29 @@
 <h3><%= title "Autoflagging Preferences" %></h3>
-<p>These are the preferences you've set for autoflagging. You can choose only certain sites to flag on by associating
-  your preferences with only those sites.</p>
-
+<h4>Getting started</h4>
+<p>
+    If you're trying to get started with autoflags for the first time, you'll need to:
+</p>
+<ul>
+  <li>authorize metasmoke for write access to your Stack Exchange account (see "Clifford the Big Red Button" below)</li>
+  <li>set up your preferences (below), which govern how many of your flags metasmoke can use per-site. You must set at least one preference.</li>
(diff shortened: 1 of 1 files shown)


thesecretmaster:
I haven't looked yet, but I have a feeling this is related to #609 and #538

ArtOfCode-:
@thesecretmaster Are you likely to have time to look into this at some point?

thesecretmaster:
@ArtOfCode Possibly, I'll try to

artofcode:
Nice ninja-edit @thesecretmaster ;-)

PR #663, opened 2019-08-19
rubocop: simple syntax changes

makyen:
This fixes some rubocop errors. These ones are adjusting whitespace and changing `CGI::unescape x` to `CGI.unescape(x)`.

In combination with the other PRs I'm submitting, it should result in CI passing. CI won't be passing in its entirety without all of the PRs being merged. This is in multiple PRs because you may not want parts of it, so only take the ones, or just the commits, which are desired.


makyen:
Closing this to re-create it in a way that results in CircleCI running on it.

PR #664, opened 2019-08-19
Quiet rubocop: these should be checked

makyen:
These are changes to quiet rubocop. I've made minimal changes to quiet each issue. However, larger changes are likely to be more appropriate. It's also possible that some of these were being used as "this should be worked on" markers.

In combination with the other PRs I'm submitting, it should result in CI passing. CI won't be passing in its entirety without all of the PRs being merged. These changes are in multiple PRs because you may not want parts of it, so only take the ones, or just the commits, which are desired. Overall the desire was to get CI to pass, so that it's easier to spot problems in future changes.

### DIFF
--- a/app/api/api/base.rb
+++ b/app/api/api/base.rb
@@ -107,7 +107,8 @@ module API
 
     before do
       key = APIKey.find_by(key: params[:key])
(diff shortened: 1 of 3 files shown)


makyen:
Closing this to re-create it in a way that results in CircleCI running on it.

PR #669, opened 2019-08-20
Test PR 01: DON'T MERGE

makyen:
This is a test to see if CircleCI runs on this commit in this repository. My expectation is that it will.

This is an alternate way to create the PR: push the branch to this repository, rather than to my fork.


makyen:
OK. This worked, as expected.

PR #673, opened 2019-08-21
rubocop issues: these should be checked

makyen:
These are changes to quiet rubocop. I've made minimal changes to quiet each issue. However, larger changes are likely to be more appropriate. It's also possible that some of these were being used as "this should be worked on" markers.

This, combined with the other PRs I'm submitting, should result in CI passing. Use the PRs, or just commits, which you want. I just feel there's considerable value to having CI normally passing, rather than always failing.

### DIFF
--- a/app/api/api/base.rb
+++ b/app/api/api/base.rb
@@ -107,7 +107,8 @@ module API
 
     before do
       key = APIKey.find_by(key: params[:key])
-      smokey_token = key.nil? ? SmokeDetector.find_by(access_token: params[:key]) : nil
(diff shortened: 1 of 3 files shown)


app/api/api/base.rb
```diff
@@ -107,7 +107,8 @@ def render(template_name, **locals)
 
     before do
       key = APIKey.find_by(key: params[:key])
```
ArtOfCode-: 109 and 110 are both lines intended for redis logging - both could be commented until redis logging can be fixed, as the auth is still done by the `authenticate_app!` call.

PR #674, opened 2019-08-21
Comment out failing tests in flagging_test.rb

makyen:
All of the tests in flagging_test.rb are failing. This PR just comments out  all of those tests. The fact that they were failing may be being used as a "this needs to be worked on" marker. If that's the case, then you may want to just close this PR.

I know that how MS handles flagging changed substantially since the last time these tests were working. I'd try to get some tests functional, but I'm just not familiar enough with the tech stack to do that in a reasonable time-frame.

Overall, my opinion is that it's better to not be trying to run these tests all the time, because it sets up the expectation that the CI tests will fail and people don't look as closely at the current CircleCI runs, because they already always fail. The fact that CI is consistently failing contributed to both myself and ArtOfCode not seeing that an earlier PR of mine caused an additional failure in the rails tests. We both missed it. If these were normally passing, then that failure (which was deployed to MS and caused the flag settings dashboard page to fail) would have almost certainly been noticed and fixed prior to that PR being merged.

In combination with the other PRs I've submitted, this should result in CI passing all three tests. This PR fixes the rails-test job, although there's still an intermittent issue with MySQL, which randomly results in some runs seeing `ConnectionError: MySQL server has gone away`.

### DIFF
--- a/test/integration/flagging_test.rb
+++ b/test/integration/flagging_test.rb
@@ -98,6 +98,7 @@ class FlaggingTest < ActionDispatch::IntegrationTest
     flag_condition.save(validate: false)
   end
 
+# rubocop:disable Style/BlockComments
 =begin
   test 'should update moderator sites' do
     @user.moderator_sites.destroy_all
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Fine for now, might be one to revisit later.

PR #676, opened 2019-08-21
nokogiri (>= 1.10.4) (security alert)

makyen:
Change nokogiri to >=1.10.4, due to the security alert [here](https://github.com/Charcoal-SE/metasmoke/network/alert/Gemfile.lock/nokogiri/open).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -232,7 +232,7 @@ GEM
       addressable (~> 2.3)
     loofah (2.2.3)
       crass (~> 1.0.2)
-      nokogiri (>= 1.5.9)
+      nokogiri (>= 1.10.4)
     mail (2.7.1)
       mini_mime (>= 0.1.1)
     marcel (0.3.3)


makyen:
Thanks for telling me. I should have noticed that from its format.

Frankly, I *really* should have been more thorough in looking at this issue. Sorry about that.

`nokogiri` is used three times in [`Gemfile.lock`](https://github.com/Charcoal-SE/metasmoke/blob/master/Gemfile.lock), stating it's a dependency for two packages: `rails-dom-testing` and `loofah`, and has a dependency itself, `mini_portile2`.  The [`Gemfile`](https://github.com/Charcoal-SE/metasmoke/blob/master/Gemfile) doesn't have an entry for `nokogiri`.

Naively, it *may* be possible to just add the `nokogiri` requirement to `Gemfile`, but we're beyond what I'd want to do without an environment in which I can build and fully test (not just the automated tests) on a machine/environment where I could run a version of the site w/o impacting other people.

At a minimum, a new `Gemfile.lock` would need to be generated through the build process. In a branch, I could shoehorn the file to be displayed by CircleCI and copy-&-paste what CircleCI generates, but I'd really want to have an actual full-up test setup to verify that there aren't related dependency issues, or that nothing actually broke through the adjustment of some related dependancy. So, doing this one correctly is beyond the environment I have set up at this time.

Basically, prior experience with dependency issues makes me hesitant to move forward at this time. So, I'm just going to close this and let one of you handle it.

thesecretmaster:
If you're OK with just updating everything (which should be fine) then you can just run `bundle update` and that will generate a new Gemfile.lock with updated dependencies.

ArtOfCode-:
I'll get to this when I've got some time at a dev machine next, no worries

PR #708, opened 2020-02-15
Change User Info page to be admin rather than developer

makyen:
The intent of this change is to make the page which summarizes user information ([example](https://metasmoke.erwaysoftware.com/users/71)) accessible to admins, not just developers. It appeared to me that the information contained in the page is user-centric, which feels more appropriate for an administrator role, rather than restricting it to developers. Most of the information is publicly available (either through the MS API, or on SE), or is directly relevant to the individual user.

The button that causes all the moderator sites to be re-determined was left as developer only, as it affects all users and is quite expensive from a SE API point of view.

As usual, these changes include a significant amount of educated guesses on my part as how all this works, so should be reviewed from that POV.

It should also be reviewed from the POV of: "should this actually be changed". This change reflects my opinion/understanding of the separation between admin and developer, which may not be consistent with what was intended.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,7 +2,8 @@
 
 class UsersController < ApplicationController
   before_action :authenticate_user!, except: [:missing_privileges]
-  before_action :verify_developer, only: %i[show refresh_ids send_password_reset update_mod_sites]
+  before_action :verify_admin, only: %i[show refresh_ids send_password_reset]
+  before_action :verify_developer, only: %i[update_mod_sites]
   before_action :set_user, only: %i[show refresh_ids send_password_reset update_mod_sites]
 
   def username; end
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
Re admin vs. developer: pretty much what they say on the tin. Admin is an administrative function. Developer is intended to be a role to grant access to the nuts and bolts and inner workings - stuff that should generally be left alone, but which requires knowledge of the internals when it does need to be poked at.

Sending password reset emails can be an admin function. However, refreshing mod sites and refreshing chat IDs should both remain as developer-only: they should generally be left well alone, but there are reasons to press them if MS isn't working properly (which requires that knowledge of how it works).

makyen:
I have no problem with `refresh_ids` being developer and have pushed a commit for it.

The main use-case I was visualizing for that button was where the MS user didn't have an account on one or more of the chat servers at the time the MS user created their MS account. At least my impression was that MS wouldn't pick up a new chat account if one was created after that point. That use-case felt more like an administrator role than developer, but can see it being developer. OTOH, I agree that any other use-case that I'm seeing for it is definitely developer.

PR #733, opened 2020-06-21
search_controller.rb: unregistered users may not use LIKE

user12986714:
In the original implementation, unregistered users' search operations on body are not redirected to fulltext index.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -47,7 +47,7 @@ class SearchController < ApplicationController
       end
     end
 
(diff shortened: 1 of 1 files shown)


NobodyNada:
Context: https://chat.stackexchange.com/transcript/message/54718795#54718795

ArtOfCode-:
Not sure I get what you're trying to do here. The restriction is deliberate - anonymous users should only be able to use the FULLTEXT search, not LIKE. This appears to drop that requirement, unless I'm missing something.

NobodyNada:
The title is a bit misleading — right now, anonymous users cannot use search *at all*, and this PR fixes that (unless user12... and I are both misunderstanding the code). Note the signed-in restriction was removed only on the `if params[:body_is_like] != '1'` branch.

From Jonathan

PR #734, opened 2020-06-22
Store post Markdown and use it as rendering source

iBug:
Got myself familiar with Rails after so many years...

Added a column to `posts` for the Markdown source fetched from API and adjusted relevant controller methods and views.

Didn't run the migration myself so `db/schema.rb` remains untouched. Surely it needs an update, too.

Side change: `params.permit` works well with Array so didn't have to expand it before passing as arguments.


ArtOfCode-:
What's the rationale behind adding Markdown storage & render, as opposed to simply rendering the HTML we already have from SE? Storing the markdown as well as the HTML has a significant cost in terms of storage that we should bear in mind. We should also consider whether our rendering is sufficient and correct - the HTML that we get from SE is the source-of-truth for what the post _should_ render as - does our rendering come up with the same HTML in all but a few edge cases?

iBug:
@ArtOfCode- We've never stored the "original" HTML that comes from SE API, but always the "processed" one, which in many cases can't render correctly (e.g. HTML in pre/code gets interpreted, not displayed as code). Storing a source makes it possible to restore "the raw look".

ArtOfCode-:
What do you mean by "processed"? We store whatever HTML Smokey sends us, which AFAIK is whatever comes from SE.

iBug:
@ArtOfCode- Sorry, that's not what comes from SE. Smokey does *a lot of* preprocessing (e.g. unescape `&lt;`, `&gt;`, `&amp;` at least) before scanning, and sends the *scanned* string instead of the raw string to MS.

ArtOfCode-:
What am I missing, @iBug? Looks like it is to me - metasmoke gets HTML from Smokey, which sends it in `Metasmoke.send_stats_on_post`, which gets passed the post body in a `Post` object from `handle_spam`. `handle_spam` is called [here](https://github.com/Charcoal-SE/SmokeDetector/blob/bd56664a7cabd6cb2178a96fc515a05a65fe0813/bodyfetcher.py#L329) and [here](https://github.com/Charcoal-SE/SmokeDetector/blob/1904f91f4469ab519fe04ac53f375cde7585b57d/chatcommands.py#L1801), both of which simply pass it the body that it gets from the API.

iBug:
@ArtOfCode- Maybe [this one](https://github.com/Charcoal-SE/SmokeDetector/blob/bd56664a7cabd6cb2178a96fc515a05a65fe0813/classes/_Post.py#L117)?

ArtOfCode-:
That just runs it through `html.unescape`, which converts HTML entities back into strings - there's no material change to the HTML in there.

makyen:
It is a material change to the HTML when the text which was converted was representing HTML text that was not intended to be rendered as HTML, but rather displayed as HTML text that's viewable to the reader. This primarily happens for `<code>` blocks which are intended to represent HTML text. In those cases, the converted text then represents HTML that is interpreted as actual HTML, elements, rather than as HTML containing code displayed as HTML text.

Unfortunately, once the HTML text is unescaped, there are some representations which can't be converted back to the actual original HTML.

ArtOfCode-:
Ah, that makes sense. Okay, lemme get the conflicts resolved...

PR #736, opened 2020-06-22
Collect all markdown renderers into one helper

iBug:
We have `WhateverController.renderer` as markdown renderers in many places, and they're backed by identical code, and they're not used for purposes other than rendering MD text.

I `grep`'d through the code and found only one case where MD rendering is called in a controller, so I grouped the `renderer.render` functions into one helper method.

In Rails 5 there's a method `helpers` available for controllers ([ref][1], not affiliated) so used it.

  [1]: https://blog.bigbinary.com/2016/06/26/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers.html


iBug:
Curious. Using class instance variables this fails (`@markdown_renderer` inside method is `nil`) while using class variables this works. Guess I have to find another way to pull this off.

makyen:
Why are we trying to Markdown-render the `post.body`? That's not Markdown. It's processed HTML, which needs it's own adjustments (not currently being performed) prior to being the actual HTML for the post (or at least as close as we can get it, as the actual HTML is not recoverable in some cases).

iBug:
Because there may be invalid HTML segments like unencoded ampersands or isolated "LT" and "GT". Passing it through a MD renderer is the easiest way to get it up and rocking.

makyen:
Yes, there is invalid HTML, but using a Markdown renderer doesn't make it display correctly. If you want JavaScript code that does that conversion, I can supply that (it's in an upcoming change to FIRE).

iBug:
That'd be another feature that'd better fit into a separate PR.

ArtOfCode-:
> Why are we trying to Markdown-render the post.body?

@makyen Probably because I wasn't thinking when I wrote that code... We could drop the `render_markdown` wrapper on that completely, and just call `raw(sanitize(post.body, ...))`.

PR #740, opened 2020-07-04
blockquote font-size: inherit, it was too big

iBug:
The default value of `17.5px` is set by Bootstrap so we can't modify it (have to override).

Problem:

> image

Solution:

> image

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -16,7 +16,7 @@ pre.post-body-pre-block {
 }
 
 blockquote {
-	font-size: inherit;
+  font-size: inherit;
 }
 
 .post-body {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/posts.scss
```diff
@@ -15,6 +15,10 @@ pre.post-body-pre-block {
   word-wrap: break-word;       /* Internet Explorer 5.5+ */
 }
 
+blockquote {
+	font-size: inherit;
```
ArtOfCode-, line 19: Can we make the indent the same 2 spaces as everything else, rather than a tab?

iBug: Oops sorry. I have `autocmd FileType css,scss set ts=2 sts=2 sw=2 et` in my editor config (tab is one less character than spaces).

PR #741, opened 2020-07-04
Replace Redcarpet with CommonMarker

iBug:
SE is fully CommonMark-ed now, so it's beneficial that we switch to a CM-compliant implementation. This should fix rendering issues with posts like [251250](https://metasmoke.erwaysoftware.com/post/251250) (autolink) and [251257](https://metasmoke.erwaysoftware.com/post/251257) (fenced code blocks).

The code is adopted from [another project of mine](https://github.com/iBug/i206hub/blob/1197fc837d962b049f3a4f69bdf3b5de291abc66/app/models/item.rb). *Battered*-tested if you'd like `:)`.

Same as my previous PRs, `Gemfile.lock` is left untouched for reasons of compatibility.

---

Actually I don't know what's wrong with Redcarpet, but [Google-ing for "redcarpet vs commonmarker"][1] gives ***all*** results on switching (instead of comparison).

  [1]: https://www.google.com/search?q=redcarpet+vs+commonmark


app/api/scripts/create_docs.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
-require 'redcarpet'
+require 'commonmarker'
 
-renderer = Redcarpet::Render::HTML.new hard_wrap: true
-parser = Redcarpet::Markdown.new renderer
+def render_markdown(text)
+  CommonMarker.render_html(text, %i[UNSAFE HARDBREAKS], %i[autolink tagfilter])
```
ArtOfCode-, line 6: [Here's a CommonMark compliant implementation](https://github.com/codidact/qpixel/blob/d77ecafed5e892b8a6cee591f1f64f34b7ab0ba8/app/helpers/application_helper.rb#L58) - this won't quite be to spec. Try this:

```suggestion
CommonMarker.render_doc(text,
                        [:LIBERAL_HTML_TAG, :STRIKETHROUGH_DOUBLE_TILDE],
                        [:strikethrough, :autolink]).to_html(:UNSAFE)
```

app/helpers/application_helper.rb
```diff
@@ -8,14 +8,21 @@ def title(text)
...
   def safe_render_markdown(text, options = {})
+    # Markdown pipeline :)
     scrubber = options[:scrubber] || PostScrubber.new
-    raw(sanitize(render_markdown(text), scrubber: scrubber))
+    s = sanitize(render_markdown(text), scrubber: scrubber)
+    doc = Nokogiri::HTML::DocumentFragment.parse s
+    doc.css('a[href]').each do |a|
+      # Like Stack Exchange does
+      a['rel'] = 'nofollow noreferrer'
+    end
```
ArtOfCode-, line 24: The Nokogiri parse is unnecessary here: `noopener noreferrer` only applies to links that also have `target="_blank"`, i.e. open in a new tab. Since Markdown links don't, you don't need to add the `rel` and thus can avoid having to parse the document (so, yay, extra performance gain).

iBug: It's not `noopener`, but `nofollow`, which is in line with that on SE. Though you're right it might be better to remove this patch altogether.

iBug:
I'm still using `CommonMarker.render_html` in `application_helpers.rb` because that helper is primarily used for rendering post previews, which IMO we should try to keep in line with what's on SE. For this reason I disabled strikethrough and tables and whatever not yet active on SE. I've read through CM's docs to make sure every option is properly understood.

PR #742, opened 2020-07-04
Use actual reason count in historical accuracy check

user12986714:
Will hopefully make db queries faster/more accurate.

### DIFF
--- a/app/models/concerns/post_concerns/autoflagging.rb
+++ b/app/models/concerns/post_concerns/autoflagging.rb
@@ -52,10 +52,13 @@ module PostConcerns::Autoflagging
         Rails.logger.warn "[autoflagging] #{id}: scaled maxes: #{scaled_maxes}"
         if !scaled_maxes.nil? && scaled_maxes.count == 6
           # Check historical accuracy
+          post_reasons_count = 0
+          post.reasons.each do |r|
+            post_reasons_count += 1
           fake_flag_condition = FlagCondition.new(
             sites: Site.mains,
             max_poster_rep: post.user_reputation,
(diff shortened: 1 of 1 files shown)


app/models/concerns/post_concerns/autoflagging.rb
```diff
@@ -55,7 +55,7 @@ def autoflag
           fake_flag_condition = FlagCondition.new(
             sites: Site.mains,
             max_poster_rep: post.user_reputation,
-            min_reason_count: 1,
+            min_reason_count: len(post.reasons),
```
ArtOfCode-, line 58: That... is not going to work. `len` is not Ruby.

app/models/concerns/post_concerns/autoflagging.rb
```diff
@@ -52,10 +52,13 @@ def autoflag
...
         if !scaled_maxes.nil? && scaled_maxes.count == 6
           # Check historical accuracy
+          post_reasons_count = 0
+          post.reasons.each do |r|
+            post_reasons_count += 1
           fake_flag_condition = FlagCondition.new(
             sites: Site.mains,
             max_poster_rep: post.user_reputation,
-            min_reason_count: 1,
+            min_reason_count: post_reasons_count,
```
ArtOfCode-, line 61: Ruby is not Python. You're writing Python-style Ruby, and... it's not going to work. You want `post.reasons.count`

user12986714: Well, I used `post.reasons.count`, and CI is not happy about it...

PR #743, opened 2020-07-05
In autoflagging, calculate the post weight outside condition validation loops

makyen:
As [discussed briefly in chat](https://chat.stackexchange.com/transcript/message/54849478#54849478), this moves the calculation of the post's weight to outside the loops which determine the users with flagging conditions which match the post.

As usual, my experience with Ruby is minimal, so please look for stupid/naive mistakes.

#### Why make this change:

Basically, `condition.validate!(post)` calls `post.reasons.pluck(:weight).reduce(:+)` every time `condition.validate!(post)` is called (i.e. each iteration through the loop). That's the way that function needs to handle it in order to validate an arbitrary post. That's good and modular OOP. However, it's inefficient when wanting to validate the same post against multiple flagging conditions.

In the loops this PR changes, we're looping through all enabled flagging conditions to validate each condition against the same post. We know we're testing the same post every loop iteration, just against a different flagging condition. So, the value returned by `post.reasons.pluck(:weight).reduce(:+)` will be the same through every iteration of the loop, but the call to `condition.validate!(post)` still results in a database access in each loop iteration to get a value which won't change.

There are, currently, 450+ existing flagging conditions.  Of those, 330+ are enabled. So, the existing code makes 330+ redundant database accesses to get the post's weight when traversing this loop, which is done for every post reported to MS. The purpose of this PR is to pull that database access out of the loop, so it's only done once, prior to the loop. The value resulting from that database access is then used in each loop iteration.

This PR should save us a few to several hundred database accesses for each post reported to MS.

### DIFF
--- a/app/models/concerns/post_concerns/autoflagging.rb
+++ b/app/models/concerns/post_concerns/autoflagging.rb
@@ -19,8 +19,11 @@ module PostConcerns::Autoflagging
       begin
         conditions = post.site.flag_conditions.where(flags_enabled: true)
         available_user_ids = {}
+        weight = post.reasons.pluck(:weight).reduce(:+)
+        poster_rep = post.stack_exchange_user.reputation
+        reasons_count = post.reasons.count
         conditions.each do |condition|
-          if condition.validate!(post)
+          if condition.validate_weight_rep_reason_count!(weight, poster_rep, reasons_count)
             available_user_ids[condition.user.id] = condition
           end
         end
@@ -167,8 +170,11 @@ module PostConcerns::Autoflagging
     def eligible_flaggers
       conditions = site.flag_conditions.where(flags_enabled: true)
       available_user_ids = {}
+      weight = reasons.pluck(:weight).reduce(:+)
+      poster_rep = stack_exchange_user.reputation
+      reasons_count = reasons.count
       conditions.each do |condition|
-        if condition.validate!(self)
+        if condition.validate_weight_rep_reason_count!(weight, poster_rep, reasons_count)
           available_user_ids[condition.user_id] = condition
         end
       end
--- a/app/models/flag_condition.rb
+++ b/app/models/flag_condition.rb
@@ -65,10 +65,14 @@ class FlagCondition < ApplicationRecord
     accuracy
   end
 
-  def validate!(post)
-    post.reasons.pluck(:weight).reduce(:+) >= min_weight &&
-      post.stack_exchange_user.reputation <= max_poster_rep &&
-      post.reasons.count >= min_reason_count
+  def validate_weight_rep_reason_count!(weight, poster_rep, reason_count)
+    weight >= min_weight && poster_rep <= max_poster_rep && reason_count >= min_reason_count
+  end
+
+  def validate_post!(post)
+    validate_weight_rep_reason_count!(post.reasons.pluck(:weight).reduce(:+),
+                                      post.stack_exchange_user.reputation,
+                                      post.reasons.count)
   end
 
   def posts
--- a/app/models/flag_condition.rb
+++ b/app/models/flag_condition.rb
@@ -69,7 +69,7 @@ class FlagCondition < ApplicationRecord
     weight >= min_weight && poster_rep <= max_poster_rep && reason_count >= min_reason_count
   end
 
-  def validate_post!(post)
+  def validate!(post)
     validate_weight_rep_reason_count!(post.reasons.pluck(:weight).reduce(:+),
                                       post.stack_exchange_user.reputation,
                                       post.reasons.count)


app/models/concerns/post_concerns/autoflagging.rb
```diff
@@ -19,8 +19,11 @@ def autoflag
       begin
         conditions = post.site.flag_conditions.where(flags_enabled: true)
         available_user_ids = {}
+        weight = post.reasons.pluck(:weight).reduce(:+)
+        poster_rep = post.stack_exchange_user.reputation
+        reasons_count = post.reasons.count
         conditions.each do |condition|
-          if condition.validate!(post)
+          if condition.validate_weight_rep_reason_count!(weight, poster_rep, reasons_count)
```
iBug, line 26: Why aren't you using `condition.validate_post! post` here?

app/models/flag_condition.rb
```diff
@@ -65,10 +65,14 @@ def self.validate_for_user(user, disabler)
...
 
-  def validate!(post)
-    post.reasons.pluck(:weight).reduce(:+) >= min_weight &&
-      post.stack_exchange_user.reputation <= max_poster_rep &&
-      post.reasons.count >= min_reason_count
+  def validate_weight_rep_reason_count!(weight, poster_rep, reason_count)
+    weight >= min_weight && poster_rep <= max_poster_rep && reason_count >= min_reason_count
+  end
+
+  def validate_post!(post)
```
ArtOfCode-, line 72: Might be a good idea to leave this one called `validate!` - just in case there are any other places in the code where `validate!` is still used.

app/models/concerns/post_concerns/autoflagging.rb
```diff
@@ -167,8 +170,11 @@ def send_not_autoflagged
     def eligible_flaggers
       conditions = site.flag_conditions.where(flags_enabled: true)
       available_user_ids = {}
+      weight = reasons.pluck(:weight).reduce(:+)
+      poster_rep = stack_exchange_user.reputation
+      reasons_count = reasons.count
       conditions.each do |condition|
-        if condition.validate!(self)
+        if condition.validate_weight_rep_reason_count!(weight, poster_rep, reasons_count)
```
iBug, line 177: Same as above, `condition.validate_post! self` is an available option.

app/models/flag_condition.rb
makyen: I'm happy to change that back. I *should have* mentioned that prior to changing it, I did a `grep "validate!" -R` on the repository. The only results were:

```
app/models/concerns/post_concerns/autoflagging.rb:23:          if condition.validate!(post)
app/models/concerns/post_concerns/autoflagging.rb:166:        if condition.validate!(self)
app/models/flag_condition.rb:68:  def validate!(post)
```

I'll change it back.

app/models/concerns/post_concerns/autoflagging.rb
makyen: The point of this PR is to not use that function there. I should have quoted the [chat conversation](https://chat.stackexchange.com/transcript/11540?m=54849478#54849478) in the PR, rather than just linking to it. I'm sorry I didn't put that context directly in the PR.

Basically, `condition.validate!(post)` calls `post.reasons.pluck(:weight).reduce(:+)` every time `condition.validate!(post)` is called (i.e. each iteration through the loop). That's the way that function needs to handle it in order to validate an arbitrary post. That's good and modular OOP. However, it's inefficient when wanting to validate the same post against multiple flagging conditions.

In the loops this PR changes, we're looping through all enabled flagging conditions to validate each condition against the same post. We know we're testing the same post every loop iteration, just against a different flagging condition. So, the value returned by `post.reasons.pluck(:weight).reduce(:+)` will be the same through every iteration of the loop, but the call to `condition.validate!(post)` still results in a database access in each loop iteration to get a value which won't change.

There are, currently, 450+ existing flagging conditions.  Of those, 330+ are enabled. So, the existing code makes 330+ redundant database accesses to get the post's weight when traversing this loop, which is done for every post reported to MS. The purpose of this PR is to pull that database access out of the loop, so it's only done once, prior to the loop. The value resulting from that database access is then used in each loop iteration.

This PR should save us a few to several hundred database accesses for each post reported to MS.

app/models/concerns/post_concerns/autoflagging.rb
makyen: Please see my prior comment.

app/models/concerns/post_concerns/autoflagging.rb
iBug: Oh yes, I should've understood this sooner.

iBug: If I were to work on this, I would have changed `Post` to add caches for these values, maybe something like this:

```ruby
class Post
  def weight
    @weight ||= reasons.pluck(:weight).reduce(:+)
  end

  def poster_rep
    @poster_rep ||= stack_exchange_user.reputation
  end

  def reason_count
    @reason_count ||= reasons.count
  end
end 
```

Then stick to `validate! post`, which is cleaner and more Ruby-ish IMO.

makyen: That's a reasonable way to do it. In fact, doing basically that was actually the first thing I thought of. I just don't know Ruby well enough, or the data structures used in MS, to be confident implementing it in that manner without *actually* testing, and I don't have an MS development environment set up in order to do that testing.

makyen: If you'd like to implement it that way and submit a new PR, I'm perfectly OK with that being the way this is resolved.

iBug: Here's my idea summed up as code:

<https://github.com/iBug/metasmoke/commit/89fb512197e90f743f0e3839f93dc79e6d65fb50>

iBug:
@makyen @ArtOfCode- I can send this branch up as a replacement PR if you find it good:

<https://github.com/iBug/metasmoke/compare/41fd0b5...ms-743>

The exact same purpose, but implemented differently (caching reason count and reason weight in the `Post` class), resulting in cleaner code and less lines changed.

makyen:
@iBug  Cool. I have no problem doing it that way. While the loops are somewhat less efficient to do that way, I'm unsure if it's enough of a performance hit to matter.

Traversing into a data structure each time to get the data is almost always going to be less compute-efficient than pulling those values out and using a local variable within the loop. However, importantly, it does make for more modular code and make any other access to `reason_weight` more efficient.

If I were doing it, I'd probably implement *both* the caching of those values and pulling the values into local variables for use in these loops. Doing so would provide both benefits.

Note: I'd probably call it `post_weight`, `reasons_weight`, `total_weight`, or `cumulative_weight` (or something similar), rather than `reason_weight`, as `reason_weight` could easily be confused with the weight of a single reason.

iBug:
#744 has been fired up for review.

BTW, I don't really think that a local variable matters *that* much in this case.

```ruby
require 'benchmark'

def test
  "hello"
end

a = Benchmark.measure do
  10000.times do
    test
  end
end
puts a
```

The output on my i7-8850H host is around 1 ms. Even taking into account that Active Record adds levels of complexity, it's unlikely we'll ever hit a practical performance issue here, as I'm already doing 10k loops and it still takes a thousandth of a second.

makyen:
@iBug I'm basically "meh" on it past eliminating the database accesses. If the code hadn't already been written, I'd probably be saying, "no, don't bother".

OTOH, I do think it's better to use a property name other than `reason_weight`, which more clearly indicates that it's the sum of the individual weights of the reasons, rather than the weight for a single reason.

makyen:
@iBug Nitpicking: your performance testing appears to be testing the performance of a function call that returns a value. It doesn't appear to be attempting to get the performance difference between referencing a local variable and accessing one that's a property of an instantiation of a class. I agree that it shouldn't be much, but there are cases, at least in other languages, where there is a notable difference, if done in time-critical code.

But, frankly, it doesn't really matter (even though, now, I have something of an itch to find out). :)

makyen:
I'm closing this in preference for the implementation in PR #744. If 744 doesn't work for some reason, feel free to reopen this one. Once 744 is merged/functional, the branch this one is on can be deleted.

PR #744, opened 2020-07-05
Implement #743 in another way

iBug:
All context is covered in #743 already.

### DIFF
--- a/app/models/flag_condition.rb
+++ b/app/models/flag_condition.rb
@@ -66,9 +66,9 @@ class FlagCondition < ApplicationRecord
   end
 
   def validate!(post)
-    post.reasons.pluck(:weight).reduce(:+) >= min_weight &&
-      post.stack_exchange_user.reputation <= max_poster_rep &&
-      post.reasons.count >= min_reason_count
+    post.reason_weight >= min_weight &&
+      post.user_reputation <= max_poster_rep &&
+      post.reason_count >= min_reason_count
   end
 
   def posts
--- a/app/models/flag_condition.rb
+++ b/app/models/flag_condition.rb
@@ -66,7 +66,7 @@ class FlagCondition < ApplicationRecord
   end
 
   def validate!(post)
-    post.reason_weight >= min_weight &&
+    post.total_weight >= min_weight &&
       post.user_reputation <= max_poster_rep &&
       post.reason_count >= min_reason_count
   end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -238,7 +238,7 @@ class Post < ApplicationRecord
   end
 
   def reason_weight
-    @weight ||= reasons.pluck(:weight).reduce(:+)
+    @weight ||= reasons.sum(:weight)
   end
 
   def user_reputation
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -241,10 +241,6 @@ class Post < ApplicationRecord
     @weight ||= reasons.sum(:weight)
   end
 
-  def user_reputation
-    @poster_rep ||= stack_exchange_user.reputation
-  end
-
   def reason_count
     @reason_count ||= reasons.count
   end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -237,6 +237,18 @@ class Post < ApplicationRecord
     native_id
   end
 
+  def reason_weight
+    @weight ||= reasons.pluck(:weight).reduce(:+)
+  end
+
+  def user_reputation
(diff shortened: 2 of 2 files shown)


makyen:
@ArtOfCode- To save you some reading on the [other PR](https://github.com/Charcoal-SE/metasmoke/pull/743): I'm good with this being the way we avoid the database accesses. It's basically how I'd first envisioned accomplishing the task, but for which I lacked the Ruby experience to be confident of implementing it without actually testing. If I had an MS dev environment set up where I could test, I probably would have gone in this direction.

So, in other words, feel free to just use this one.

ArtOfCode-:
⚠️ Has this been tested? It may not do what you think it does.

This approach will reload each of these attributes directly from the database every time a new model instance is created. I don't know enough about the internals of ActiveRecord to tell you exactly when that happens, but it's not something that's in our control and could potentially happen... basically whenever - particularly around model changes.

thesecretmaster:
@ArtOfCode- Are you sure about that? That doesn't sound right to me. Are you talking about `#total_weight` and `#reason_count`? If so, those should be recalc'd only when those methods are called on an instance of the model that they haven't been called on before, unless my understanding of AR is very wrong.

ArtOfCode-:
That's what I mean, @thesecretmaster - so every time you get a new model instance and call those methods, they recalc from the DB. If I knew more about how AR works internally, I'd be able to tell if that's a problem or not, but since I don't... as long as this is tested and confirmed working, we'll go with it.

thesecretmaster:
AR won't call these instance methods internally, they should only be called if we call them. I can go ahead and test this is my newly set up dev env for sanity, but 99% sure it will be fine.

thesecretmaster:
Yeah, sanity confirmed:

```
irb(main):006:0> post = Post.find(12345)
  Post Load (2.3ms)  SELECT  `posts`.* FROM `posts` WHERE `posts`.`id` = 12345 LIMIT 1
=> #<Post id: 12345, title: "http://naturalsmaleenhancement.com/brainplus-iq/", body: "I've never read a detail on Brainplus Iq. That onl...", link: "//askubuntu.com/questions/714118", post_creation_date: nil, created_at: "2015-12-28 07:20:51", updated_at: "2016-02-23 16:55:14", site_id: 29, user_link: "//askubuntu.com/users/486769/darlenesanchez", username: "darlenesanchez", why: nil, user_reputation: nil, score: nil, upvote_count: nil, downvote_count: nil, stack_exchange_user_id: 8251, is_tp: true, is_fp: false, is_naa: false, revision_count: nil, deleted_at: nil, smoke_detector_id: nil, autoflagged: false, tags: nil, feedbacks_count: 2, native_id: 714118, markdown: nil>
irb(main):007:0> post.total_weight
   (3.8ms)  SELECT SUM(`reasons`.`weight`) FROM `reasons` INNER JOIN `posts_reasons` ON `reasons`.`id` = `posts_reasons`.`reason_id` WHERE `posts_reasons`.`post_id` = 12345
=> 394
irb(main):008:0> post.total_weight
=> 394
irb(main):009:0> post.reason_count
   (0.4ms)  SELECT COUNT(*) FROM `reasons` INNER JOIN `posts_reasons` ON `reasons`.`id` = `posts_reasons`.`reason_id` WHERE `posts_reasons`.`post_id` = 12345
=> 4
irb(main):010:0> post.reason_count
=> 4
```

makyen:
While it's a low frequency event, what happens when the total weight of the post is 0? I'm assuming it will get re-fetched from the database each time `.total_weight` is read.

A post having 0 weight happens rarely enough such that we can probably accept the hit, or the test to prevent recalc could be that the value is undefined/not yet assigned.

iBug:
@makyen It's not going to happen. In Ruby only `nil` and `false` are falsey (failing `if`, passing `unless`, and causing `||` to select the value on the right), everything else is truth, so a value of `0` won't cause the RHS of `||=` to be evaluated.

iBug:
@thesecretmaster You might not have understood Art correctly. He probably meant something like this:

```ruby
a = Post.find(12345)
b = Post.find(12345)
a.total_weight  # A DB query
b.total_weight  # Another DB query
a.total_weight == b.total_weight  # Both cached this time
```

Even though `a` and `b` refers to the same object, they're different Ruby instances, and thus have different caching setup.

I admit this is how it would actually work, but given that at present we're only using these two methods in flag condition validation, it's far from being a problem.

makyen:
@iBug Thanks. I should have looked up falsey values for Ruby. I guess I'm too used to JavaScript & C. :;

PR #745, opened 2020-07-06
Clean up old code in flagging method

thesecretmaster:
I just stumbled across this confusing being/rescue statement and felt the need to do some cleanup. Putting it in a PR since the code is in an important place and I don't want to break things.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -195,10 +195,10 @@ class User < ApplicationRecord
       spam: ['spam', 'contenido no deseado', 'スパム', 'спам'],
       abusive: ['rude or abusive', 'rude ou abusivo', 'irrespetuoso o abusivo', '失礼又は暴言', 'невежливый или оскорбительный'],
       other: ['in need of moderator intervention', 'precisa de atenção dos moderadores', 'se necesita la intervención de un moderador',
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -183,31 +183,28 @@ def flag(flag_type, post, dry_run = false, **opts)
...
+      abusive: ['rude or abusive', 'rude ou abusivo', 'irrespetuoso o abusivo', '失礼又は暴言', 'невежливый или оскорбительный'],
+      other: ['in need of moderator intervention', 'precisa de atenção dos moderadores', 'se necesita la intervención de un moderador',
+              'モデレーターによる対応が必要です', 'требуется вмешательство модератора']
+    }
+
+    unless flag_strings.keys.include? flag_type.to_sym
+      return false, "Unrecognized flag type #{flag_type} specified in call to User#flag"
+    end
+
+    flag_option = flag_options.find do |fo|
```
ArtOfCode-, line 205: `find`?

thesecretmaster: Like select, but stops after the first time it finds one. From [Enumerable](https://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-find).

ArtOfCode-: Huh, TIL

PR #753, opened 2020-07-13
Added the ability to remove user added domains

thesecretmaster:
This does exactly what the title says, continuing my work from https://github.com/Charcoal-SE/metasmoke/commit/0b2d47e4aba25b32f84aa4a542e8440d6765501c. The only problem is that I needed to add an attribute on the join table `posts_spam_domains`, but when I created a model for that table everything got angry about a compound primary key and the lack of an `id` column. I don't know what the best solution is here.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -28,7 +28,7 @@ class Post < ApplicationRecord
   has_and_belongs_to_many :reasons
   has_and_belongs_to_many :post_tags, class_name: 'DomainTag'
   has_many :post_spam_domains
-  has_and_belongs_to_many :spam_domains
+  has_many :spam_domains, through: :post_spam_domains
   has_many :feedbacks, dependent: :destroy
   has_many :deletion_logs, dependent: :destroy
   has_many :flag_logs, dependent: :destroy
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -27,6 +27,7 @@ class Post < ApplicationRecord
   belongs_to :smoke_detector
   has_and_belongs_to_many :reasons
   has_and_belongs_to_many :post_tags, class_name: 'DomainTag'
+  has_many :post_spam_domains
   has_and_belongs_to_many :spam_domains
   has_many :feedbacks, dependent: :destroy
   has_many :deletion_logs, dependent: :destroy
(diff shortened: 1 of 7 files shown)


app/models/post.rb
```diff
@@ -27,6 +27,7 @@ class Post < ApplicationRecord
   belongs_to :smoke_detector
   has_and_belongs_to_many :reasons
   has_and_belongs_to_many :post_tags, class_name: 'DomainTag'
+  has_many :post_spam_domains
   has_and_belongs_to_many :spam_domains
```
ArtOfCode-, line 31: Should be a `has_many :through` now that there's a model on the join too

app/models/spam_domain.rb
```diff
@@ -4,6 +4,7 @@ class SpamDomain < ApplicationRecord
   include Websocket
 
   has_and_belongs_to_many :posts, after_add: :setup_review
```
ArtOfCode-, line 6: Likewise, `has_many :through`

app/views/posts/show.html.erb
```diff
@@ -187,7 +187,15 @@
 <% if @post.spam_domains.any? %>
   <ul class="post-domain-list">
     <% @post.spam_domains.each do |d| %>
-      <li><%= render 'spam_domains/domain', domain: d, occurances: true %></li>
+      <li>
+        <%= render 'spam_domains/domain', domain: d, occurances: true %>
+        <% if current_user&.has_role?(:core) && !(psd = PostSpamDomain.find_by(post: @post, spam_domain: d)).added_by.nil? %>
```
ArtOfCode-, line 192: Declaring variables inside conditional statements is... not wonderful. The declaration should move to the line above and just be used in the conditional.

thesecretmaster: I actually really like this style :( I'm pretty sure rubocop is OK with it as well.

app/models/spam_domain.rb
thesecretmaster: Can do, the only reason I didn't the first time is because I didn't want rails to be angry at me for changing things. It was already angry about the composite primary key

app/views/posts/show.html.erb
thesecretmaster: Not fixing for now because I want to get this in before the deploy, but let me know if you feel strongly about it and we can fix it on master.

PR #758, opened 2020-07-13
Actually restrict db dumps to core users

user12986714:
Currently it appears that anyone with the correct url can obtain the db and redis dump.

### DIFF
--- a/app/controllers/dumps_controller.rb
+++ b/app/controllers/dumps_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DumpsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :verify_core
   before_action :set_dump, only: %i[show edit update destroy]
 
   # GET /dumps


thesecretmaster:
Pretty sure this is status-bydesign, although feel free to correct me if someone knows better.

user12986714:
@thesecretmaster https://charcoal-se.org/smokey/Privileges#core-ms
```
Ability to download database dumps*
```

thesecretmaster:
Ah right, IIRC this had something to do with S3 usage/costs (cc @Undo1?) If that's still true than we should make it visible to Core users in the navbar (see `app/views/layouts/application.html.erb`), if that's not true post-server-migration then I think we should leave it open to everyone and correct the wiki.

teward:
We currently disabled backup dumps for DB and such so having the URL won't do anything for now.  Redis dumps aren't regularly available as far as I can tell, even when we were running daily DB backups.

Further, the DB dumps are uploaded to S3 and only in sanitized form (no PII is leaked in them) which is good for statistical analysis and other analysis people want to do on the data.  And because PII is stripped from the SQL dumps it's a non-issue if non-core people have access since they can just ask for the distribution anyways:

> * Although downloading database dumps is restricted to core users, you’re welcome to distribute them onwards to other people who ask for them too.

Certain ones of us with God Access to the machine have full access to the database though, including the un-sanitized dumps - this is for backup purposes and for recovery purposes, or for certain levels of testing such as trying to debug DB level problems or DB index issues or such.

This is by-design and overall not really a 'concern' since the sanitized dumps can be distributed freely if someone asks someone with Core access to pull a dump.  It's the private DB dumps we have to worry about sharing, and those only reside in heavily-secured backups I maintain currently in the infrastructure.

thesecretmaster:
I think the reason for the core + distribution thing is that the dumps were rather large and we didn't want to rack up Undo's S3 bill. So I think we'd stick with core only + distributing if they'll still be hosted on S3.

teward:
> Also cc @teward about where DB dumps will be hosted once they're back up? Will they still be S3 or will they be hosted off the same box as MS?

Currently, there are no DB dumps available because of me still building out the replacement to Metasmoke and Ruby trying to DB dump and breaking things.  If we want we can host the single files off of the Metasmoke instance itself, or we can spin S3 style hosting dumps in my own environment with S3 compatible storage infrastructure.  The only place we'd keep multiple backups is for the unsanitized DB storage in my own highly-secure backup systems for recovery purposes - I keep at least weekly backups in storage right now in the prototype system for DB backups)

Given that I have zero actual data bandwidth caps we should be able to host the backups right from MS and serve direct (or run it via S3 compatible storage locally).

However, in either of the last two statements, we don't *need* to enforce the Core requirement because there's no cost for storage in my infrastructure other than my power bill.

user12986714:
SD privileges wiki has been changed to reflect this.

teward:
> I think the reason for the core + distribution thing is that the dumps were rather large and we didn't want to rack up Undo's S3 bill. So I think we'd stick with core only + distributing if they'll still be hosted on S3.

Yep, you're right, this was because of S3 costs, per Undo:

> The reason before was to keep someone we've irritated from downloading it nine million times and ballooning the S3 bandwidth out cost.

With us moving to on-server copies of the DB and sharing them directly from the Metasmoke server, plus a few dozen security controls on the network border, we no longer need to enforce Core restrictions for DB access.

PR #761, opened 2020-07-13
Feedback "rude" is also positive

user12986714:
As seen in
```ruby
  VALID_TYPES = %w[tp tpu fp fpu naa rude ignore]
                .map { |f| [f, "#{f}-"] }.flatten
```

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -123,7 +123,7 @@ class Feedback < ApplicationRecord
   end
 
   def is_positive? # rubocop:disable Style/PredicateName
-    feedback_type.include? 't'
+    feedback_type.include? 't' || feedback_type.include? 'rude'
   end
 
   def is_negative? # rubocop:disable Style/PredicateName
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -123,7 +123,7 @@ class Feedback < ApplicationRecord
   end
 
   def is_positive? # rubocop:disable Style/PredicateName
-    feedback_type.include? 't' || feedback_type.include? 'rude'
+    feedback_type.include?('t') || feedback_type.include?('rude')
   end
 
   def is_negative? # rubocop:disable Style/PredicateName


thesecretmaster:
Not sure we want to add rude to the `is_positive?` check. Have you looked into where we call `is_positive?` and checked if we want to be including rude in those places?

makyen:
At least in SD, any of `spam`, `rude`, `abusive`, or `offensive` are treated as the equivalent to `tpu-` (see: [Privileged commands as reply](https://github.com/Charcoal-SE/SmokeDetector/wiki/Commands#privileged-commands-as-reply)). I haven't double checked the code, but I thought those got translated to `tpu-` prior to being sent to MS. It's also possible we might have some of these alternate feedbacks from feedback submitted to the MS API.

I'd be interested to see how many feedbacks in MS this actually affects. If it's significant for `rude`, then it may be an issue for the other explicit possibilities. Another solution might be to just go through the database and change all of the equivalent alternates to straight `tpu-`.

user12986714:
Just checked all (according to grep) code and the answer appears to be yes - we want to include `rude` in `is_positive()`.

user12986714:
An alternative might be to remove `rude` feedback altogether, and do a db migration that changes them to `tpu`

thesecretmaster:
To me, it looks like we want the tp only behavior, at least here:

https://github.com/Charcoal-SE/metasmoke/blob/053c482dd3f580f692f99c83fecb79019f29c5b4/app/models/post.rb#L184

user12986714:
Why would we want to exclude `rude` from `is_tp` attribute of `post`? `is_tp` is parallel with `is_fp` and `is_naa`.

ArtOfCode-:
```
MariaDB [metasmoke]> select distinct feedback_type from feedbacks;
+---------------------------------------------------------------------------+
| feedback_type                                                             |
+---------------------------------------------------------------------------+
| tpu                                                                       |
| fpu                                                                       |
| tpu-                                                                      |
| fp                                                                        |
| tp                                                                        |
| fp-                                                                       |
| tp-                                                                       |
| fpu-                                                                      |
| ignore-                                                                   |
| ignore                                                                    |
| ignore-_shouty_case                                                       |
| trueu-                                                                    |
| trueu                                                                     |
| naa                                                                       |
| invalid                                                                   |
| naa-                                                                      |
| false                                                                     |
| naa2                                                                      |
| true                                                                      |
| Mith_now_youve_made_everyone_do_it                                        |
| This feature is brought to you                                            |
| This is Smokey. I've come to say goodbye, as I've been permanently broken |
+---------------------------------------------------------------------------+
```

IOW: `rude` is translated by Smokey into `tpu-` before it gets sent to metasmoke; metasmoke has no record of `rude` as a feedback.

PR #762, opened 2020-07-14
Restrict domain link destroy to admin

user12986714:


ArtOfCode-:
This is intentionally a core privilege

PR #763, opened 2020-07-14
Restrict add_review to admin

user12986714:
Arbitrary user can blow up review queue otherwise


ArtOfCode-:
What's the intention of this? I think you may be misunderstanding.

user12986714:
Not valid

PR #765, opened 2020-07-14
Don't double nullify flag logs

user12986714:
`flag_logs` is already associated with
```ruby
has_many :flag_logs, dependent: :nullify
```


user12986714:
Might be the cause of losing flag logs.

user12986714:
I thought the foreign key constraint is removed on d845bd91 ...

user12986714:
Using `unscoped` on associated model will cause the return of all model **regardless the association**, effectively nuking all flag logs.

thesecretmaster:
I'm not quite sure why we're removing the foreign key, but removing that before_destroy callback looks good to me.

user12986714:
@thesecretmaster Otherwise we can't delete user due to foreign key constraints; see CI log for the first commit.

thesecretmaster:
Yeah, but shouldn't the flag logs user_id field get nullified before the deletion, making it a non-issue?

user12986714:
@thesecretmaster But... CI does not agree with this; looks like it is nullified *after* the deletion of the db row in table Users, which will cause the action of deleting the Users row to fail.

thesecretmaster:
Yeah, but I think the fix should be to troubleshoot the order in which those things happen, not remove the constraint. There's nothing wrong with the constraint as far as I can tell.

ArtOfCode-:
Eh, Rails fudges FK constraints anyway

PR #766, opened 2020-07-14
Prevent shell injection on grep

user12986714:
This effectively gives any developer UNRESTRICTED SHELL ACCESS.

### DIFF
--- a/app/controllers/developer_controller.rb
+++ b/app/controllers/developer_controller.rb
@@ -13,7 +13,10 @@ class DeveloperController < ApplicationController
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+           unless params[:grep].instance_of?(String) && params[:context].instance_of?(Integer)
+             raise ActionController::RoutingError, 'Not Found'
+           end
+             `tail -n 10000 log/production.log | grep -E '#{params[:grep].tr("'", '')}' -C '#{params[:context]}' --color=never`
            else
              `tail -n 1000 log/production.log`
            end


Undo1:
This is mostly by-design; any developer probably has SSH access anyway. The tr here might be useful, though I suspect there are other ways to inject stuff in if you want to.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,10 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             unless params[:grep].instance_of?(String) && params[:context].instance_of?(Integer)
```
thesecretmaster, line 16: Have you tested this? I generally prefer to avoid `instance_of?` checks and instead try to attempt a coercion (`to_s` or `to_i`. Also, this seems like it would break grep search since I don't see you converting `context` to an integer anywhere and I believe params are all strings by default.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,18 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             begin
+               grep_str = params[:grep].instance_of?(String) ? params[:grep] : params[:grep].to_s
+               context_int = params[:context].instance_of?(Integer) ? params[:context] : params[:context].to_i
```
thesecretmaster, line 18: You can just run `params[:grep].to_s`. `params[:grep]` will always be either nil or a string. Same is true for `params[:context]`.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,18 @@ def update_sites
...
+               # Check if coercion is successful, in case someone overloaded .to_*() methods
+               if !grep_str.instance_of?(String) || !context_int.instance_of?(Integer)
+                 raise ArgumentError, 'Coercion failure'
```
thesecretmaster, line 21: Raising in a rescue is a very weird thing to do and is almost never a good thing, especially since rescuing only re-raises a different error.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,18 @@ def update_sites
...
+               end
+             rescue
+               # Coercion failure -> deny access
+               raise ActionController::RoutingError, 'Not Found'
```
thesecretmaster, line 25: This is the wrong error to raise, I'd probably `flash[:warning]` or something instead.

app/controllers/developer_controller.rb
user12986714: The intention of `rescue` is mainly to catch `Missing method` errors raised by `.to_*()`, as we probably can't assume `params` is a dictionary of strings (may have nested dictionaries, for example); the `raise ArgumentError` is a reuse of the `rescue` statement in an effort to make the code less redundant.

thesecretmaster: If you want to catch Method Missing errors, use `respond_to?` instead. It's a much more standard way of checking if an object responds to a method.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             grep_str = if params[:grep].instance_of?(String)
```
thesecretmaster, line 16: This part of the `if` statement is redundant, since Strings respond to to_s.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+                          params[:grep]
+                        elsif params[:grep].respond_to?(to_s)
```
thesecretmaster, line 18: You need a colon before `to_s`

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+                          params[:grep].to_s
+                        else
+                          flash[:warning] = 'Coercion failure on params[:grep]'
+                          redirect_to(root_path) && return
+                        end
```
thesecretmaster, line 23: You also don't need any of this error handling at all, since `params[:grep]` is always going to be a string, hash, or array, all of which respond to `to_s`, and in fact will automatically be coerced to a string in the string interpolation on line 37.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+             context_int = if params[:context].instance_of?(Integer)
```
thesecretmaster, line 24: This will always be false, `params[:context]` will always be a string, hash, or array.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+                             params[:context]
+                           elsif params[:context].respond_to?(to_i)
```
thesecretmaster, line 26: Colon before `:to_i`

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+                             params[:context].to_i
+                           else
+                             flash[:warning] = 'Coercion failure on params[:context]'
+                             redirect_to(root_path) && return
+                           end
+             # Ensure are of correct types in case .to_*() get overrided.
+             unless grep_str.instance_of?(String) && context_int.instance_of?(Integer)
```
thesecretmaster, line 33: If you're doing all of that error checking above, you don't need to duplicate it here

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,28 @@ def update_sites
...
+                        end
+             context_int = if params[:context].instance_of?(Integer)
+                             params[:context]
+                           elsif params[:context].respond_to?(to_i)
+                             params[:context].to_i
+                           else
+                             flash[:warning] = 'Coercion failure on params[:context]'
+                             redirect_to(root_path) && return
+                           end
+             # Ensure are of correct types in case .to_*() get overrided.
```
thesecretmaster, line 32: There is no sane world in which we would override these methods to not return the appropriate type. There's no need to check this.

app/controllers/developer_controller.rb
thesecretmaster: I actually am very curious about how this didn't make CI fail with a NameError. What you're doing is equivalent to `params[:context].respond_to?(to_i())`, which is to say that you're passing the return value of the `to_i` method to `respond_to?`. Now, there's no reason why the `to_i` method should be defined within an instance of ApplicationController, the only reason I can imagine this not just throwing an error is because rails does some weird `method_missing` magic which I kind of want to dig in to, but I gotta run so maybe later.

thesecretmaster:
Also feel free to ping me in chat if you have questions with any of my comments, I feel like we're kinda going in circles here.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,14 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             grep_str = params[:grep].to_s
+             context_int = if params[:context].respond_to?(:to_i)
+                             params[:context].to_i
+                           else
+                             flash[:warning] = 'Coercion failure on params[:context]'
```
thesecretmaster, line 20: I'd use a more informative error and instead say "Context was not a number" or something like that.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,14 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             grep_str = params[:grep].to_s
```
thesecretmaster, line 16: This is very minor, but I'd personally prefer to inline this (`params[:grep].to_s.tr("'", '')`)

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,14 @@ def update_sites
...
+             context_int = if params[:context].respond_to?(:to_i)
```
thesecretmaster, line 17: Also very minor, but this might look better if written as a guard instead, so

```ruby
unless params[:context].respond_to?(:to_i)
    # flash, redirect, return
end
```

and then in the interpolation use `params[:context].to_i`.

thesecretmaster:
Ah shoot that was supposed to be "general comment" not "request changes" -- none of those changes are mandatory if you don't like them.

app/controllers/developer_controller.rb
```diff
@@ -13,7 +13,11 @@ def update_sites
 
   def production_log
     @log = if params[:grep].present?
-             `tail -n 10000 log/production.log | grep -E '#{params[:grep]}' -C #{params[:context]} --color=never`
+             unless params[:context].respond_to?(:to_i)
+               flash[:warning] = 'Coercion failure on params[:context]: Can\'t convert to an interger'
```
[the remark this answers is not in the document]
user12986714: If there is ever a failure, something is badly wrong, and we probably want devs to know where to check/fix

PR #767, opened 2020-07-14
Authenticate user before signing up for email

user12986714:
Not particularly sure what `emails` do; it just does not look right to access current_user without authentication.


thesecretmaster:
I don't know devise well enough to know if this is actually necessary or not, but either way this definitely isn't doing any harm.

PR #768, opened 2020-07-14
Actually prevent unregistered users from LIKE searches

user12986714:
The original implementation only displays a flash, but still allows users to perform LIKE searches.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -55,6 +55,7 @@ class SearchController < ApplicationController
       else
         if params[:body_is_like] == '1' && !user_signed_in?
           flash[:warning] = 'Unregistered users cannot use LIKE searches on the body field. Please sign in.'
+          redirect_to(root_path) && return
         end
         # Otherwise, it's REGEX or NOT REGEX, which fulltext won't do - fall back on search_string and params
         search_string << "IFNULL(`posts`.`body`, '') #{body_operation} :body"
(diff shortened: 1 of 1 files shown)


app/controllers/search_controller.rb
```diff
@@ -55,6 +55,7 @@ def index
       else
         if params[:body_is_like] == '1' && !user_signed_in?
           flash[:warning] = 'Unregistered users cannot use LIKE searches on the body field. Please sign in.'
+          redirect_to(root_path) && return
```
thesecretmaster, line 58: You don't need a `return` statement, `redirect_to` automatically terminates the request.

user12986714:
@thesecretmaster [I don't think so](https://api.rubyonrails.org/classes/ActionController/Redirecting.html)
```
Statements after redirect_to in our controller get executed, so redirect_to doesn't stop the execution of the function. To terminate the execution of the function immediately after the redirect_to, use return.
```

thesecretmaster:
Oh huh TIL. I have no idea how I've gone this long with that misconception :/

thesecretmaster:
I'd prefer to redirect users to /search instead of / though, otherwise looks good to me. However search code is a bit messy and it isn't my mess so I don't feel comfortable approving it.

user12986714:
@ArtOfCode- It appears that I can't request a review/add a reviewer for some reasons, so...
```
[Wed Jul 15 03:21 UTC 2020] Requested a review from ArtOfCode-
```

PR #771, opened 2020-07-15
Raise exception on protect_from_forgery

user12986714:
Make sure the session hard fails.

### DIFF
--- a/app/controllers/custom_sessions_controller.rb
+++ b/app/controllers/custom_sessions_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class CustomSessionsController < Devise::SessionsController
-  protect_from_forgery except: [:create]
+  protect_from_forgery except: :create, with: :exception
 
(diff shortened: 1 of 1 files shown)


thesecretmaster:
Is there a problem with the default, null_session approach?

user12986714:
That .verify_code() method looks susceptible to request forging; however I am not quite sure what the :null_session really does & if it is secure

ArtOfCode-:
`null_session` drops the session cookie, i.e. logs the user out... but that's not the default approach; `exception` is. This is already protected by exception.

thesecretmaster:
@ArtOfCode- [Are you sure?](https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L134)

ArtOfCode-:
@thesecretmaster Pretty sure; I seem to remember the Rails docs saying `null_session` is the default for API apps, but `exception` is the default for web apps.

PR #772, opened 2020-07-15
protect_from_forgery

user12986714:
Especially the .destroy() method

### DIFF
--- a/app/controllers/custom_registrations_controller.rb
+++ b/app/controllers/custom_registrations_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CustomRegistrationsController < Devise::RegistrationsController
+  protect_from_forgery with: :exception
+
   def create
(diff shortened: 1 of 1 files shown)


ArtOfCode-:
Have you tested that this works/doesn't break anything?

user12986714:
Nope, but it *should* be good since no one should be creating/deleting a user without an authenticated session; but without testing I am not quite sure either.

ArtOfCode-:
Let me just check: do you understand what `protect_from_forgery` is doing?

user12986714:
It will add a `_csrf_token` and validate against it, so to make sure the session is not forged.

ArtOfCode-:
What does "the session is not forged" mean, though? It sounds like you have a misunderstanding of the attack vector this is protecting against - have a look at what CSRF is.

user12986714:
Not valid

PR #776, opened 2020-07-16
Fix fix asn tags (hopefully)

user12986714:
Hopefully fixes the NoMethodError


app/models/spam_domain.rb
```diff
@@ -20,7 +20,7 @@ def fix_asn_tags
     asn_query = `dig +short "$(dig +short '#{domain.tr("'", '')}' | awk -F. '/[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/ {print $4"."$3"." $2"."$1;exit}').origin.asn.cymru.com" TXT` # rubocop:disable Metrics/LineLength
     asn = asn_query.strip.tr('"', '').split('|')[0]&.strip
     return unless asn.present?
-    prev_domain_tags = domain_tags.where(special: true).select(:name).map(&:name).select { |dt| dt.start_with?('AS-') }.map(&:name)
+    prev_domain_tags = domain_tags.where(special: true).select(:name).select { |dt| dt.start_with?('AS-') }.map(&:name)
```
ArtOfCode-, line 23: This isn't going to work. `select(:name)` will return a collection of ActiveRecord objects, which don't have a `start_with?` method because they're not strings.

PR #780, opened 2020-07-19
Check if conflict before removing posts from review

user12986714:
We won't generally want to remove a post from the review queue when the feedback is `['tpu', 'fp']`.


user12986714:
This has gone completely bit rot. I am no longer able to understand what I wrote and cannot fix the CI.

PR #781, opened 2020-07-20
Prevent non-reviewers from submitting feedbacks

user12986714:


Undo1:
Good catch

Undo1:
Trivial rubocop failure there on the `if` that can be an `unless`.

user12986714:
I actually pushed a commit *after* this get merged :-P to fix rubocop...

Undo1:
Throw another PR and I'll merge it

PR #782, opened 2020-07-20
Fix rubocop attempt

user12986714:
Replace `if !(blah)` with `unless blah`


user12986714:
@Undo1 It is here...

Undo1:
Danke

PR #783, opened 2020-07-21
Auto review job

user12986714:
Automatically review all posts older than 1 month, have no true positive feedback and is not deleted as 'fp-'.

This will hopefully reduce the amount of posts in the review queue.

### DIFF
--- a/app/jobs/auto_review_job.rb
+++ b/app/jobs/auto_review_job.rb
@@ -2,7 +2,7 @@
 
 class AutoReviewJob < ApplicationJob
   queue_as :default
-  
+
   def perform
     Rails.logger.info "Job started (#{DateTime.now})"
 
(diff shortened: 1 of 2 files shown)


makyen:
No, let's not do this. This just introduces potentially bad data. I'd rather see the post records deleted than potentially add erroneous feedback. Note: I don't want the posts deleted. While the criteria you are using makes it more likely that the post is FP, it *certainly* doesn't guarantee it.

Undo1:
It's a good thought, and would be perfect in a lot of applications, but we value correct data a lot (helps with community trust, etc.) It's better for us to have a big queue and good data than a small queue and questionable data (even if it's right, it's still *questionable* from the outside).

teward:
I'm with Makyen and Undo in this.  Autoreviews like this will lead to bad data, and that's bad for *our* system.

user12986714:
Well, I may be missing something, but I think this means "we are lowering the requirement to 1 manual fp feedback for old and undeleted posts" rather than "we are not going to review them at all"?

PR #787, opened 2020-07-31
Disallow users to supply redirect parameter

user12986714:
Not really a security issue, but is not-so-good practice anyway.


ArtOfCode-:
[status-declined] [status-bydesign]

This is explicitly by-design to allow server-side apps to pass a redirect parameter when requesting OAuth, so that they can get the token as soon as the user has authorized without requiring the extra step that client-side apps must go through.

PR #790, opened 2020-08-13
Add global post hashmap to speed up by_url method

user12986714:
It is known that by using id database fetch will be *much* faster.

I am not sure why we are using by_uid method at all though.


ArtOfCode-:
[status-declined] - both site and native ID are indexed, so there's no speed difference between ID or UID, meaning this adds a lot of code for almost zero gain.

PR #793, opened 2020-08-18
Suffix tree for basic pattern match

user12986714:
Implemented suffix tree for basic pattern match (i.e. exact matches).
More functionalities may be implemented in the future to expand to regex
or approximate matches.

**Notes**  
1) Due to unfamiliarity with Rails, this implementation may be erroneous or contain various issues.
2) Please do read the C code before merge even if it looks terrible. Skipping this step means that we are running untrusted code on the server as no one can say for sure that I am not malicious.
3) Please do not merge unilaterally; discussions is likely needed.


user12986714:
Things to consider:
1. How to migrate post data to the tree? The current implementation is to create a blank tree as ActiveRecord::Migration and perform async migration by having developers manually doing so with st_insert_post_range, or POST dev/st_add_range . If we are going to do that, *do not* add a large range; instead, add multiple small ranges to be extra careful. To ensure performance, ranges should be added sequentially w.r.t. post id, that is, for example, first 1 - 1000 and then 1001 - 2001. Note that the insertion is fast at first but will slow down as more post get added.
2. I am directly responsible for the code and hence the bugs. However, without more access I really cannot do much in case an error occurs. On the other hand, it is a terrible idea to give me dev privilege just for this.
3. Currently this functionality is restricted to core. This is to prevent malicious exploitation of bugs just in case. At some time, this need to go to other users. However, there is a DoS risk by using a reader preference rwlock in C code. More discussions are needed prior to open this up to unregistered users.

user12986714:
I am not sure if there is a security risk not sanitizing "reason" string provided by users in st_mark_broken().

thesecretmaster:
Couple of questions: 1) What's the disk overhead of this going to look like? 2) Where are the files going to live / how many files will this be creating?

user12986714:
@thesecretmaster For the first, I am not really sure but should be between 10~100GiB. For the second, exactly one file (this may be changed prior to this PR being merged)

thesecretmaster:
Gonna go ahead and tag @teward here, since this is significantly larger than the entire database. Loading a database dump into my dev environment gives me ~600mb (at least according to according to [a magic query](https://stackoverflow.com/questions/1733507/how-to-get-size-of-mysql-database)), and checking the actual size of the database on disk give me only 1.2gb.

user12986714:
Actually, most code review should probably go to https://github.com/user12986714/SuffixTree/blob/master/ext/suffix_tree/suffix_tree.c

thesecretmaster:
Yeah I saw, and I think I'm going to end up being the person who's going to review it, but I want to make sure it's practical before I put the time in to reviewing all that

user12986714:
@thesecretmaster If so, maybe not to review the C code for another 2 days. I almost had my experiments done and a major update improving data locality/disk usage will be pushed soon.

teward:
@thesecretmaster @user12986714 if you are not 100% sure of the code yet or still have major improvements in the pipeline, this should be a Draft PR not a "ready for merge" PR.  My two cents.

teward:
> Gonna go ahead and tag @teward here, since this is significantly larger than the entire database. Loading a database dump into my dev environment gives me ~600mb (at least according to according to [a magic query](https://stackoverflow.com/questions/1733507/how-to-get-size-of-mysql-database)), and checking the actual size of the database on disk gives me only 1.2gb. 

I have to take a look, the DB is not huge on disk, but there is likely indexes and such taking space too.

user12986714:
@thesecretmaster (cc @NobodyNada ) The work-in-process refactoring is at user12986714/SuffixTree:dev-refactor . It is still a work in process, but should finished very soon as the core part has been done and only some wrappers are needed.

thesecretmaster:
I'm going to hold off on reviewing the C extension until we get the go-ahead from Thomas that this will actually see the light of day on prod. Also, could you add some toggle somewhere that either drastically reduces disk usage (to under ~1gb or less) or disables all this suffix tree stuff entirely so that we don't have to have MS take up a ton of disk when we're just developing locally?

@teward: Is adding 10-100gb of files of suffix tree data feasible? Does the prod instance even have that much disk space?

teward:
@thesecretmaster we currently are using ~20GB out of a 245G partition (250GB disk, of which 5GB is allocated to swap).  On the SSD that Metasmoke is sort of dedicated to, there's expansion space for at least 200GB more if it's absolutely needed.  (Disk space is not something that's at a premium right now, if that's what you're asking.)

However, as this *is* an SSD, I'm a little concerned about the I/O overhead not just disk space overhead.  What's the resource overhead such as disk IO and CPU needs for this, not just the disk space?

(Also ~10GB - 100GB is a **wide** range which means you haven't actually *audited* this to determine what the actual disk impact would be.  Which means if I extrapolate this 'range' for a larger dataset, it can go anywhere up to TBs of data for extra overhead).

teward:
@user12986714 My primary question is:

> Implemented suffix tree for basic pattern match (i.e. exact matches).

How does this improve existing mechanisms?  Is there a specific use case that is so heavily demanded that it *justifies* (at least) quadrupling the disk overhead for the DB, etc?

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+  def self.available_fields
+    FIELD_BITMASK.keys
+  end
+
+  def self.calc_mask(fields)
+    mask = 0
+    fields.each do |f|
+      mask |= FIELD_BITMASK[f]
+    end
+    mask
```
thesecretmaster, line 49: `#inject` might be a little cleaner here: `fields.inject(0) { |mask, field| mask | FIELD_BITMASK[field] }`

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  def self.basic_search(pattern, mask)
+    SuffixTreeSingleton.instance.basic_search(pattern.unicode_normalize.encode('utf-8'), mask)
+  end
+
+  def self.insert_post(post_id)
+    post = Post.find post_id
+    post.nil? && (raise ArgumentError)
```
thesecretmaster, line 58: `raise ArgumentError if post.nil?`

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+    body: 2,
+    why: 4,
+    username: 8
+  }.freeze
+
+  def self.available_fields
+    FIELD_BITMASK.keys
+  end
+
+  def self.calc_mask(fields)
```
thesecretmaster, line 44: You'll prolly want to intersect `fields` with `available_fields` to avoid `nil` related errors

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+    SuffixTreeSingleton.instance.mark_broken reason
+  end
+
+  def self.functional?
+    SuffixTreeSingleton.instance.functional?
+  end
+
+  def self.broken_reason
+    SuffixTreeSingleton.instance.broken_reason
+  end
```
thesecretmaster, line 83: This could be made simpler with either `Forwardable` from stdlib, `delegate` from rails, or a simple loop:
```
%i[sync! mark_functional mark_broken ....].each do |method|
  define_singleton_method(method) { SuffixTreeSingleton.instance.send(method) }
end
```

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def self.basic_search(pattern, mask)
+    SuffixTreeSingleton.instance.basic_search(pattern.unicode_normalize.encode('utf-8'), mask)
+  end
+
+  def self.insert_post(post_id)
+    post = Post.find post_id
+    post.nil? && (raise ArgumentError)
+    available_fields.each do |f|
+      SuffixTreeSingleton.instance.insert(post.send(f).unicode_normalize.encode('utf-8'), FIELD_BITMASK[f], post.id)
```
thesecretmaster, line 60: Use `post[f]` to avoid accidentally calling methods that you don't want to call.

user12986714:
@teward The growth is O(n) to the size of all posts; that is, if we have doubled our dataset, we will double the disk usage. I am not particularly sure about the actual constant before the n as that is going to depend on the characteristics of all those posts (LCP and other things). The thought is that this change will hopefully make search *much* faster.

The recent refactor tries to improve data locality, so I hope that disk IO issue is somewhat reduced. CPU usage is almost never an issue compared to (relatively) slow disk IOs.

app/helpers/suffix_tree_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+      @reason = reason
+    end
+
+    def functional?
+      @functional
+    end
+
+    def broken_reason
+      @reason
+    end
```
thesecretmaster, line 30: I'd recommend just setting `attr_accessor :broken_reason` and then calling that accessor from `#mark_broken` and `#initialize` (although you don't actually need to initialize it, if you don't it'll just be `nil` which shouldn't cause any problems that I can see)

app/controllers/search_controller.rb
```diff
@@ -319,4 +323,130 @@ def index_fast
...
+      end
+      format.rss { render :search, layout: false }
+      format.xml { render 'search.rss', layout: false }
+    end
+  end
+
+  private
+
+  def check_st_functional
+    SuffixTreeHelper.functional? && return
```
thesecretmaster, line 449: `return if SuffixTreeHelper.functional?`

thesecretmaster: Or also `render ... unless SuffixTreeHelper.functional?`

app/controllers/search_controller.rb
```diff
@@ -319,4 +323,130 @@ def index_fast
...
+        when 'naa'
+          @results = @results.where(is_naa: true)
+        end
+
+        @sites = Site.where(id: @results.map(&:site_id)).to_a unless params[:option] == 'graphs'
+        render :search
+      end
+      format.json do
+        render json: @results
+      end
```
thesecretmaster, line 440: Why `do ... end` here and braces below?

user12986714:
@thesecretmaster W.r.t. disabling the suffix tree thing: this is the whole point of this PR. The idea is to trade space for time. A suffix tree based search is O(m) complexity while a BM (algorithm used by LIKE) based is O(n). This means that when we have more and more posts, BM will take more and more time (currently already at ~10s) while suffix tree will take constant time linear to the length of the pattern to search.

app/controllers/developer_controller.rb
```diff
@@ -3,8 +3,57 @@
...
+    render plain: 'Success', status: 200
+  end
+
+  def st_insert_post_range
+    BatchInsertPostToSuffixTreeJob.perform_later((params[:start_id].to_i..params[:end_id].to_i))
+    render plain: 'Job started asynchronously', status: 200
+  end
+
+  def st_basic_search_raw
+    @post_ids = SuffixTreeHelper.basic_search(params[:pattern], params[:mask])
```
thesecretmaster, line 43: I don't think you need this to be an instance variable

app/jobs/insert_post_to_suffix_tree_job.rb
```diff
@@ -0,0 +1,12 @@
...
+class InsertPostToSuffixTreeJob < ApplicationJob
+  queue_as :default
+
+  def perform(post_id)
+    Rails.logger.info "Job started (#{DateTime.now}, +0s)"
+    SuffixTreeHelper.insert_post post_id
+    SuffixTreeHelper.sync_async
+    Rails.logger.info "Job finished (#{DateTime.now})"
+  end
+end
```
thesecretmaster, line 12: You don't need this job, just alter the batch job to take args as a splat with `def perform(*post_ids)` and then `post_ids = post_ids.flatten` if you want to be extra safe but I don't think you need that.

app/jobs/batch_insert_post_to_suffix_tree_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class BatchInsertPostToSuffixTreeJob < ApplicationJob
```
thesecretmaster, line 3: I like `InsertPostsToSuffixTreeJob` better but that's just me

thesecretmaster:
Yeah, I'm not saying it should be shrunken / disabled on prod, I just want a flag so I don't have to deal with all that disk usage in my dev environment :P

teward:
> The growth is O(n) to the size of all posts;

That's very concerning, because while we're at 20GB total disk usage including OS right now, if we need 100GB for the current amount of posts and then another 100GB when we get even MORE reports in, that's **not scaleable** a feasible manner , because eventually (probably sooner than later) we'd exhaust the existing memory storage of the existing SSD, unless people want to start donating funds regularly to fund newer hardware / storage media for the cluster (or to identify 'dedicated hardware' for Metasmoke and the ancillary support system which does backups right now).

Even if @thesecretmaster approves the changes, **I will require Undo and/or ArtOfCode's support to deploy this**.  Because I currently believe it's not scaleable over time.

teward:
> BM (algorithm used by LIKE) based is O(n)

Except this algorithm in use by LIKE and such at the DB level doesn't have any *disk overhead* impact.  So the question is, do we really want to trade off disk space for a somewhat faster 'like' behavior.  Personally, I would say "no" because O(n) on **disk space per post** is not scaleable.

Undo1:
Agree - this is cool tech, but if it's not scaleable it's not scaleable. Further, deploying something that is highly dependent on a single person is fraught with peril when personnel changes happen (or the person who understands it has Real Life happen to them, etc.)

Maintenance issues could be addressed, scaleability issues will depend on the constant - this'll be fundamentally Ω(cn) in space at least. If c can be very small, scaleability may not be an issue. If it's large, I can't support doing something that locks us in to having loads of disk space and IO requirements.

user12986714:
Looks like there is something wrong with the extension; I am still investigating.
Basically, the rewrite caused catastrophic disk usage compared to the one prior to the rewrite.

teward:
> Looks like there is something wrong with the extension; I am still investigating.
> Basically, the rewrite caused catastrophic disk usage compared to the one prior to the rewrite.

Even non catastrophic disk usage is still not scalable.  Unless the disk-cost-per-post is really small we will basically not going to be easily able to scale this out, as Undo stated in his reply

user12986714:
@teward We do not know. My personal estimation (once the issue get fixed, which will be soon) is about ~30 GiB for all posts, which is not *terribly* disk demanding. (When it grows to 300 GiB, BM will take 100 seconds on a search)

teward:
> @teward We do not know. My personal estimation (once the issue get fixed, which will be soon) is about ~30 GiB for all posts, which is not _terribly_ disk demanding. (When it grows to 300 GiB, BM will take 100 seconds on a search)

Alright, so the 'irritated sysadmin' in me has been awoken.  Congrats, because when that side of me wakes up, it's usually a dark day filled with me ranting and complaining.

I actually went **and did some research into disk size and such**.  The combined total of data in `/var/lib/mysql` - including an old  dump db and an old temp DB comes to a total of 4GB of data.

```root@metasmoke:/var/lib/mysql# du -h --max-depth=1 . | sort -hr
4.0G	.
1.6G	./metasmoke_temp
1.6G	./metasmoke
687M	./dump_metasmoke
6.6M	./mysql
8.0K	./performance_schema```

The actual *data* that we have for just the Metasmoke database is a 1.6GB data set.  Your estimation is ~30GB for a 1.6GB dataset, which means that from a *file storage perspective* your disk demand overhead for a 1.6GB dataset is **18.75 times** of the original size of the dataset larger.

Taking that value and extrapolating this multiplier (18.75 times the disk set size) let's say the database expands to 300GB as you suggest.  Disk overhead requirements at that point for your tree will be **5.625TB** just for a 300GB MySQL dataset.  And that's not including the 300GB of the database, so that total disk requisite comes out to just shy of 6TB (5.925 TB).

That doesn't scale at the "we're generally volunteers and don't get paid to maintain this infrastructure" level that we're at.  Sure, my infrastructure is funded through my LLC and my own business but that doesn't mean I'm paid nearly enough (thanks to the COVID crisis) to continually sink ten grand into brand new cutting edge hardware every year.


> Edit: Concerning the scalability issue, the asymptotic complexity is already optimal. The issue is to reduce the constant.

Also, congrats on making me have to dredge out one of my older comp-sci textbooks on this one - 'asymptotic complexity' was not in my regular vocabulary so I had to brush up a tad briefly before digging and responding.

**Comment by user12986714** [comment]
At the time this comment is posted, there are [65,997](https://metasmoke.erwaysoftware.com/data/sql/queries/214-number-of-posts-in-the-last-year) posts created within 365 days. To double the data size, we need another three years.

When the suffix tree disk usage reaches 300 GiB, if my estimation of 30 GiB is accurate, we need 27 years.  
That is a looong time. I am not sure if the project is going to exist for that long.  

Concerning the 5.6 TiB issue: when the main database reaches 300 GiB, it has been expanded 187 times. To "achieve" that, we need 558 years (:-P). At the same time, won't you rather have a 5.6 TiB data set than having to wait 1870 seconds for every search?

**Comment by teward** [comment]
> Concerning the 5.6 TiB issue: when the main database reaches 300 GiB, it has been expanded 187 times. To "achieve" that, we need 558 years (:-P). At the same time, won't you rather have a 5.6 TiB data set than having to wait 1870 seconds for every search?

Current needs vs. future needs.  You're looking at user friendliness and search speed expediency.  I'm looking at the disk cost, cost to maintain that storage array, cost to *expand* the storage array, and because we're near max SATA slots in SSD data on the compute nodes, compute node costs (~$1.4k to get a new server added to the compute cluster)

So unless we want to go back to spinning disk mediums (which is what my SAN contains currently) this is just not *cost* effective.

Unless you want to donate a new compute node to the cluster.  Happy to accept a ~$1.4k donation of hardware.

**Comment by teward** [comment]
So, right now i'm cloning MS so we have a 'test' one we can run this against for actual ideas of disk storage (up to 210GB total) and determine the actual multiplier.  Because the 30GB overhead is estimated on a 10% sample size (20k posts) and not the full set.

Once MS is cloned, even if it's not up to date we can get a better idea of the disk overhead requirements on 220k posts instead of 20k posts.  (Closer to 'ready for review' time that is).

File: app/helpers/suffix_tree_helper.rb (hunk 3 above)
Comment (user12986714) [reply app/helpers/suffix_tree_helper.rb:44]: I *think* I have already done the intersection before calling calc_mask...

**Comment by user12986714** [comment]
The bug is now fixed.

Concerning reducing the constant, I have some ideas that can almost halve the disk usage (or even more!). However, this will reduce the extensibility of the gem. It is possible for me to create a Charcoal-SE clone just for this, while keeping the generic user12986714/SuffixTree repo extensible.

File: app/controllers/developer_controller.rb [hunk 12]
```diff
@@ -3,8 +3,57 @@
...
   before_action :check_impersonating, only: %i[change_back verify_elevation]
 
+  def st_mark_functional
+    Rails.logger.warn "Suffix tree extension marked functional by #{current_user.username}"
+    SuffixTreeHelper.mark_functional
+    render plain: 'Success', status: 200
+  end
+
+  def st_mark_broken
+    reason = if params.key?(:reason) && !params[:reason].empty?
```
thesecretmaster, line 17: `params[:reason].present?` to avoid `undefined method empty? on nil`

app/controllers/search_controller.rb
```diff
@@ -319,4 +323,128 @@ def index_fast
...
+                 Post.all.includes_for_post_row
+               end
+
+    per_page = user_signed_in? && params[:per_page].present? ? [params[:per_page].to_i, 10_000].min : 100
+
+    if params.key?(:pattern) && !params[:pattern].blank?
+      filters = SuffixTreeHelper.available_fields.filter do |f|
+        params.key?(f)
+      end
+      filters.empty? && filters = SuffixTreeHelper.available_fields
```
thesecretmaster, line 349: filters = ... if filters.empty?

app/helpers/suffix_tree_helper.rb
thesecretmaster: My feeling is that it makes more logical sense to do that here, every single caller of this method will want the intersect and so that logic should live in this method.

app/jobs/sync_suffix_tree_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class SyncSuffixTreeJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Rails.logger.info "Job started (#{DateTime.now}, +0s)"
```
thesecretmaster, line 7: No need to long start and finish timestamps, rails already logs that.

config/config.sample.yml
```diff
@@ -243,13 +243,23 @@ common: &common
 
 development:
   <<: *common
+  suffix_tree:
+    str_path: './dev_str_pool'
+    tag_path: './dev_tag_pool'
+    child_path: './dev_child_pool'
+    node_path: './dev_node_pool'
```
thesecretmaster, line 250: If these files are going to exist, don't forget to add them to .gitignore

config/routes.rb
```diff
@@ -193,6 +194,14 @@
...
 
   scope 'dev' do
+    post 'st_functional',    to: 'developer#st_mark_functional'
+    post 'st_broken',        to: 'developer#st_mark_broken'
+    post 'st_add_post',      to: 'developer#st_insert_post'
+    post 'st_add_post_sync', to: 'developer#st_insert_post_synchronous'
+    post 'st_add_range',     to: 'developer#st_insert_post_range'
+    get  'st_basic_search',  to: 'developer#st_basic_search_raw'
+    post 'st_sync',          to: 'developer#st_sync'
+    post 'st_async_sync',    to: 'developer#st_async_sync'
```
thesecretmaster, line 204: All the dev routes (and search_st) should end up in the nav or at least linked from somewhere

app/controllers/developer_controller.rb
```diff
@@ -14,7 +14,7 @@ def st_mark_functional
   end
 
   def st_mark_broken
-    reason = if params.key?(:reason) && !params[:reason].empty?
+    reason = if params[:reason].present? && !params[:reason].empty?
```
thesecretmaster, line 17: The empty test is redundant, `present?` only returns true if !nil? && !empty?

user12986714:
I just did an experiment which reveals that I am doing too much space-time trade off; I have came up a way to dramatically reduce disk usage while keeping search reasonably fast.

config/routes.rb
user12986714: I have absolutely zero knowledge/experience in UI design or front end dev. I am not sure how to do this... Maybe make a big "Suffix tree control panel"?

thesecretmaster: You should be able to add them in /app/views/layouts/application.html.erb, that's where the navbar gets built.

user12986714:
As it currently stands, it is unlikely that this PR will do what it is intended. Therefore, I am closing it as we have much more important issues to spend development time on.  

---
#### What this PR was trying to solve  
Basically, metasmoke search is quite slow for both exact string matches and regex matches. Furthermore, current search algorithms takes linear or hyper-linear time with respect to the length of text to be searched. Hence, we can expect search to be slower as we get more posts.  
This PR implemented a suffix trie in an attempt to solve such problems. Theoretically, a search on a suffix trie takes linear time with respect to the length of *the pattern to be searched*, which is much less.  
#### Why it didn't work  
I failed to take into consideration that, in fact, the time complexity of trie based search is `O(m + n)`, with `m` being the length of the pattern string and `n` be the number of matches. It is often the case that `n` is either negligibly small, or it is required to return all the match (thus `O(n)` time is required anyway). Unfortunately, our application is not such a case. We need to return all the document containing at least one match, but not to return every match.  
There is a way (that I am aware of) which can turn `n` into the number of documents to return, solving the time problem. However, it will lead to exponential storage requirement, which is not sustainable given our current hardware resources and capacities.

PR #794, opened 2020-08-20
Not to unilaterally nuke posts

user12986714:
Currently, the spam wave tool allows posts matching certain regex to be flagged with up to 6 flags. This does not look good, as we can make mistakes and spam wave flags are not subject to user's flag conditions. This PR changes the value to 5 so that no post is unilaterally nuked by autoflags.

### DIFF
--- a/app/models/spam_wave.rb
+++ b/app/models/spam_wave.rb
@@ -7,7 +7,7 @@ class SpamWave < ApplicationRecord
   serialize :conditions, JSON
 
   validates :name, presence: true
-  validates :max_flags, numericality: { greater_than_or_equal_to: 1, less_than_or_equal_to: 6 }
+  validates :max_flags, numericality: { greater_than_or_equal_to: 1, less_than_or_equal_to: 5 }
   validate :check_min_accuracy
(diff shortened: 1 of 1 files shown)


makyen:
At the time the spam tool was created, the choice was specifically to allow for the the possibility of 6 flags being selected.

I don't see anything that has changed which would affect the decision to allow for that possibility.

What happens if, after consultation with the moderators of a site or with the CMs, the moderators of that site want, or the CMs desire, for six flags to be raised on a post?

What if there is a spammer, or group of spammers who are posting 150+ posts/hour and people can't keep up? (that *has* happened) *And* site mods/CMs want us to use 6 flags?

Other than to assuage the feelings of some people, none of whom have I seen mention this specific thing as an issue, I don't see a reason to do this.

On the other hand, if you wanted to add a confirmation dialog to verify that it really was the intent to use 6 flags, I would not object to that.

Note: If you are wanting to actually prevent a spam wave from nuking a post, then you need to also account for sites where it requires only 3 flags to red-flag-delete a post.

PR #799, opened 2020-08-26
Validate against months parameter

user12986714:
It appears that site dashboard will have erroneous behaviour and take relatively long time to run when a large negative "months" parameter is supplied (example: `-10000000000000000000000000000000000000000`).

This PR validates months parameter and ensures that it is not less than 1, which is the value "enforced" in the corresponding javascript.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -74,7 +74,7 @@ class DashboardController < ApplicationController
     @site = Site.find(params[:site_id])
 
     @months = params[:months].to_s.empty? ? 3 : params[:months].to_i
-    @months < 1 && @months = 1
+    @months = 1 if @months < 1
     @months_string = @months == 1 ? 'month' : "#{@months} months"
 
     @all_posts = @posts.where(site_id: @site.id)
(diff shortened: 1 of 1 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -74,7 +74,8 @@ def site_dash
     @site = Site.find(params[:site_id])
 
     @months = params[:months].to_s.empty? ? 3 : params[:months].to_i
-    @months_string = @months <= 1 ? 'month' : "#{@months} months"
+    @months < 1 && @months = 1
```
ArtOfCode-, line 77: wat-face.jpg

If this is meant to be a conditional assignment, make it an actual conditional assignment (`x = y if z == 1`, or with a full if statement)

ArtOfCode- (changes requested):

app/controllers/dashboard_controller.rb
user12986714: I think rubocop is known to complain about the full if statement and pointed me to the && one, but looks like it is fine with `... if ...` construct.

thesecretmaster:
https://api.rubyonrails.org/v6.0.3.2/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize

_(yes, I know I wrote this code, I was a fool)_

PR #801, opened 2020-08-30
Fix fix asn tags... again

user12986714:
We need to first delete tags before adding new ones to prevent duplicated tags error.

### DIFF
--- a/app/models/spam_domain.rb
+++ b/app/models/spam_domain.rb
@@ -20,14 +20,19 @@ class SpamDomain < ApplicationRecord
     asn_query = `dig +short "$(dig +short '#{domain.tr("'", '')}' | awk -F. '/[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/ {print $4"."$3"." $2"."$1;exit}').origin.asn.cymru.com" TXT` # rubocop:disable Metrics/LineLength
     asn = asn_query.strip.tr('"', '').split('|')[0]&.strip
     return unless asn.present?
-    domain_tags.where(special: true).select(:name).map(&:name).select { |dt| dt.start_with?('AS-') }.each do |dtn|
-      domain_tags.delete(DomainTag.find_by(name: dtn))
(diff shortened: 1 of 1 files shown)


thesecretmaster:
What problem is this solving? In what case is there a duplicated tags error?

user12986714:
POST https://metasmoke.erwaysoftware.com/domains/3/fix_asn , gives 
```
ActiveRecord::RecordNotUnique in SpamDomainsController#fix_asn_tags
Mysql2::Error: Duplicate entry '323-3' for key 'PRIMARY': INSERT INTO `domain_tags_spam_domains` (`domain_tag_id`, `spam_domain_id`) VALUES (323, 3)
```

app/models/spam_domain.rb
```diff
@@ -20,17 +20,15 @@ def fix_asn_tags
...
-    prev_domain_tags = domain_tags.where(special: true).select(:name).map(&:name).select { |dt| dt.start_with?('AS-') }
+    domain_tags.where(special: true).select(:name).map(&:name).select { |dt| dt.start_with?('AS-') }.each do |dtn|
+      domain_tags.delete(DomainTag.find_by(name: dtn))
+    end
     asn.split.each do |as|
       desc = `dig +short AS#{as}.asn.cymru.com TXT`.strip.tr('"', '').split('|')[-1]&.strip
       tag = DomainTag.find_or_create_by(name: "AS-#{as}", special: true)
-      prev_domain_tags -= ["AS-#{as}"]
       tag.update(description: "Domains under the Autonomous System Number #{as} - #{desc}.") unless tag.description.present?
       domain_tags << tag
```
thesecretmaster, line 30: I'd prefer to fix this by sticking an `unless domain_tags.include? tag` on this line.

thesecretmaster: It's just going to be slightly better than clearing all the AS tags then re-creating them.

PR #806, opened 2020-09-06
Remove dead code

user12986714:
This PR tries to remove code that is no longer in use.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -2,7 +2,6 @@
 
 class APIController < ApplicationController
   before_action :verify_key, except: %i[filter_generator api_docs filter_fields calculate_filter]
-  before_action :verify_trusted_key, only: [:regex_search]
   before_action :set_pagesize, except: %i[filter_generator api_docs]
   before_action :verify_write_token, only: %i[create_feedback report_post spam_flag add_domain_tag]
   skip_before_action :verify_authenticity_token, only: %i[posts_by_url create_feedback report_post spam_flag post_deleted add_domain_tag
--- a/app/controllers/dumps_controller.rb
+++ b/app/controllers/dumps_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class DumpsController < ApplicationController
-  before_action :set_dump, only: %i[show edit update destroy]
-
   # GET /dumps
   # GET /dumps.json
   def index
--- a/app/controllers/smoke_detectors_controller.rb
+++ b/app/controllers/smoke_detectors_controller.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class SmokeDetectorsController < ApplicationController
-  before_action :authenticate_user!, except: %i[audits check_token]
-  before_action :verify_admin, except: %i[audits force_failover force_pull mine token_regen new create check_token]
+  before_action :authenticate_user!, except: :audits
+  before_action :verify_admin, except: %i[audits force_failover force_pull mine token_regen new create]
   before_action :verify_blacklist_manager, only: %i[force_failover force_pull]
   before_action :verify_smoke_detector_runner, only: %i[mine token_regen new create]
-  before_action :set_smoke_detector, except: %i[audits mine new create check_token]
+  before_action :set_smoke_detector, except: %i[audits mine new create]
 
   def destroy
     unless current_user.present? && (current_user.has_role?(:admin) || current_user.id == @smoke_detector.user_id)
@@ -81,20 +81,6 @@ class SmokeDetectorsController < ApplicationController
     end
 
     redirect_to params[:redirect] || smoke_detector_mine_path
-  end
-
-  # Used by Helios to verify new tokens
-  def check_token
-    token = SmokeDetector.where(access_token: params[:token]).first
-    payload = {
-      exists: token.present?,
-      owner_name: token&.user&.username,
-      location: token&.location,
-      created_at: token&.created_at,
-      updated_at: token&.updated_at
-    }
-
-    render json: payload
   end
 
   private
--- a/app/views/users/migrate_token_confirmation.html.erb
+++ b/app/views/users/migrate_token_confirmation.html.erb
@@ -1,4 +1,0 @@
-<%= form_tag migrate_token_path do %>
-  <%= hidden_field_tag "state", params[:state] %>
-  <%= submit_tag "Confirm that you've granted access", class: 'btn btn-primary' %>
-<% end %>
--- a/db/migrate/20200906202319_drop_table_dumps.rb
+++ b/db/migrate/20200906202319_drop_table_dumps.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class DropTableDumps < ActiveRecord::Migration[5.2]
+  execute 'DROP TABLE dumps'
+end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -132,27 +132,6 @@ class UsersController < ApplicationController
     redirect_to dev_user_path(@user)
   end
 
-  def migrate_token_confirmation
-    return unless params[:state] != Rails.cache.fetch("token_migration_state/#{current_user.id}")
-    flash[:danger] = "It looks like you didn't actually authenticate! It might still work, but I don't think it will."
-  end
-
-  def migrate_token
-    state = Rails.cache.fetch("token_migration_state/#{current_user.id}")
-    res = HTTParty.get("#{AppConfig['token_store']['host']}/auth/confirm",
-                       headers: { 'X-API-Key': AppConfig['token_store']['key'] },
-                       query: { account_id: current_user.stack_exchange_account_id, state: state })
-    Rails.logger.info res
-    if params[:state] == state && JSON.parse(res.body)['token_exists'] && current_user.update(write_authenticated: true)
-      flash[:success] = 'Your registration was completed sucessfully!'
-    elsif params[:error].present?
-      flash[:danger] = "Got an error: #{params[:error]}: #{params[:error_description]}"
-    else
-      flash[:danger] = "We couldn't complete your registration. Please try again."
-    end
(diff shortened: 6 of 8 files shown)


app/controllers/smoke_detectors_controller.rb
```diff
@@ -83,20 +83,6 @@ def token_regen
...
-    payload = {
-      exists: token.present?,
-      owner_name: token&.user&.username,
-      location: token&.location,
-      created_at: token&.created_at,
-      updated_at: token&.updated_at
-    }
-
-    render json: payload
-  end
```
thesecretmaster, line 98: I forget when this was introduced, but I think we will want to keep this unless helios is official dead.

app/views/dashboard/index.html.erb
```diff
@@ -1,9 +1,3 @@
-<form role="search">
-  <div class="form-group">
-    <input type="text" id="search" class="form-control" placeholder="Search reasons">
-  </div>
-</form>
-
```
thesecretmaster, line 6: This should still be functional / not dead?

thesecretmaster (commented):
I'd like to also keep some of the dumps code, although it's dead now, it may be alive again if we move off of Thomas's infra.

app/views/dashboard/index.html.erb
user12986714: No, this never has worked... It just displays a search box but it does not actually do anything (i.e. won't function at all)

user12986714:
> I'd like to also keep some of the dumps code, although it's dead now, it may be alive again if we move off of Thomas's infra.

Well, the point being the code is originally bugged for performance and security, so if we move away from the infra we have to rewrite them anyway,

app/views/dashboard/index.html.erb
thesecretmaster: I've heard that complaint from others, but it seems to work on prod on my machine so I'd like to leave it.

app/controllers/api_controller.rb
```diff
@@ -2,7 +2,6 @@
 
 class APIController < ApplicationController
   before_action :verify_key, except: %i[filter_generator api_docs filter_fields calculate_filter]
-  before_action :verify_trusted_key, only: [:regex_search]
```
ArtOfCode-, line 5: Why is this dead?

ArtOfCode-:
A lot of this seems like stuff that's potentially still alive/useful. Could you add a comment to each of the files changed to explain what the rationale for considering it dead is?

app/controllers/api_controller.rb
user12986714, line 5: This method has been deleted long before.

app/controllers/dumps_controller.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class DumpsController < ApplicationController
-  before_action :set_dump, only: %i[show edit update destroy]
-
```
user12986714, line 5: The four method in only clause does not exist.

app/controllers/smoke_detectors_controller.rb
```diff
@@ -1,11 +1,11 @@
...
 
 class SmokeDetectorsController < ApplicationController
-  before_action :authenticate_user!, except: %i[audits check_token]
-  before_action :verify_admin, except: %i[audits force_failover force_pull mine token_regen new create check_token]
+  before_action :authenticate_user!, except: :audits
+  before_action :verify_admin, except: %i[audits force_failover force_pull mine token_regen new create]
   before_action :verify_blacklist_manager, only: %i[force_failover force_pull]
   before_action :verify_smoke_detector_runner, only: %i[mine token_regen new create]
-  before_action :set_smoke_detector, except: %i[audits mine new create check_token]
+  before_action :set_smoke_detector, except: %i[audits mine new create]
```
user12986714, line 8: I don't think Helios is still active but this may need to be reverted.

app/javascript/admin.js
```diff
@@ -45,25 +45,6 @@ onLoad(() => {
     });
   });
 
-  $('input.pin-checkbox').change(function () {
```
user12986714, line 48: Pinned roles are indistinguishable from non-pinned roles.

app/models/application_record.rb
```diff
@@ -3,11 +3,6 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
-  def self.fuzzy_search(term, **cols)
-    sanitized = sanitize_for_search term, **cols
-    select(Arel.sql("`#{table_name}`.*, #{sanitized} AS search_score"))
-  end
-
```
user12986714, line 10: This method is never used.

app/models/application_record.rb
```diff
@@ -37,15 +32,6 @@ def self.sanitize_for_search(term, **cols)
...
 
-  # From http://stackoverflow.com/questions/6591722/how-to-generate-fixtures-based-on-my-development-database
-  def dump_fixture
-    fixture_file = "#{Rails.root}/test/fixtures/#{self.class.table_name}.yml"
-    File.open(fixture_file, 'a') do |f|
-      f.puts({ "#{self.class.table_name.singularize}_#{id}" => attributes }
-        .to_yaml.sub!(/---\s?/, "\n"))
-    end
-  end
-
```
user12986714, line 48: This method is no longer being used.

app/models/application_record.rb
```diff
@@ -69,14 +55,4 @@ def self.mass_habtm(join_table, first_type, second_type, record_pairs)
...
-  def self.full_dump
-    username = Rails.configuration.database_configuration[Rails.env]['username']
-    password = Rails.configuration.database_configuration[Rails.env]['password']
-    host = Rails.configuration.database_configuration[Rails.env]['host']
-    `#{Rails.root}/dump/dump.sh "#{username}" "#{password}" "#{host}"`
-
-    Dump.destroy_all
-    Dump.create file: File.open(Dir.glob('dumps/*')[0])
-  end
```
user12986714, line 81: Bugged and no longer used.

app/models/post.rb
```diff
@@ -230,10 +230,6 @@ def deleted?
     deletion_logs.where(is_deleted: true).any?
   end
 
-  def conflicted?
-    is_tp && (is_fp || is_naa)
-  end
-
```
user12986714, line 236: This method is never used.

app/models/reason.rb
```diff
@@ -39,9 +39,4 @@ def both_percentage
 
     count.to_f / posts.count
   end
-
-  # Attempt to use cached post_count if it's available (included in the dashboard/index query)
-  def fast_post_count
-    try(:post_count) || posts.count
-  end
```
user12986714, line 46: This method is never used.

app/models/user.rb
```diff
@@ -238,10 +238,6 @@ def spam_flag(post, dry_run = false)
     flag :spam, post, dry_run
   end
 
-  def abusive_flag(post, dry_run = false)
-    flag :abusive, post, dry_run
-  end
-
```
user12986714, line 244: This method is never used.

app/models/user.rb
```diff
@@ -250,20 +246,6 @@ def moderator?
...
-    else
-      UsersRole.create(user: self, role: role, pinned: true)
-    end
-  end
-
-  # rubocop:disable Style/PredicateName
-  def has_pinned_role?(role)
-    UsersRole.where(user: self, role: Role.find_by(name: role), pinned: true).exists?
-  end
-
```
user12986714, line 266: Indistinguishable

app/views/users/migrate_token_confirmation.html.erb
```diff
@@ -1,4 +0,0 @@
-<%= form_tag migrate_token_path do %>
-  <%= hidden_field_tag "state", params[:state] %>
-  <%= submit_tag "Confirm that you've granted access", class: 'btn btn-primary' %>
-<% end %>
```
user12986714, line 4: This file is no longer being linked to, and I think all token migration is finished.

db/migrate/20200906202319_drop_table_dumps.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class DropTableDumps < ActiveRecord::Migration[5.2]
+  def change
+    execute 'DROP TABLE dumps'
+  end
```
user12986714, line 6: Table no longer being used after Rails based full dump being disabled.

db/migrate/20200906202320_remove_api_token_legacy.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class RemoveAPITokenLegacy < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :users, :encrypted_api_token_legacy
+    remove_column :users, :token_migrated_legacy
```
user12986714, line 6: Legacy column; may introduce security risks.

app/models/application_record.rb
ArtOfCode-: Ah - this one does need to stay. It's not used in normal operation, no, but it's there to provide easy seeding for development.

app/models/post.rb
ArtOfCode-: Worth keeping around, though - potentially useful and not doing any harm.

app/models/user.rb
ArtOfCode-: No harm in keeping it, though

ArtOfCode- (changes requested):

app/models/user.rb
thesecretmaster: Are we sure that this works after the token-store migration?

thesecretmaster:
This all looks good to me, but I want to get confirmation as to Helios being officially dead yet. I feel like it's still a thing we want to make happen eventually?

user12986714:
> This all looks good to me, but I want to get confirmation as to Helios being officially dead yet. I feel like it's still a thing we want to make happen eventually?

Well, Helios is archived and is way out dated. It is highly unlikely that the original project will become active again. Although we do want that kind of change, it is likely that we need to rewrite it.

Issue #792, opened 2020-08-18
Can still see exact review count


user12986714:
![image](https://user-images.githubusercontent.com/65436504/90456357-d29ef280-e0c6-11ea-8b11-461e289e7d35.png)

It appears that one can see actual review count (instead of '50+') after submitting a feedback on a post, if and only if the post originally has no feedback.

user12986714:
Step to reproduction: select a post with no feedback. Submit a feedback by clicking on either of the button. Now '50+' suddenly becomes the actual number (in this case 24590)

thesecretmaster:
Closing as a duplicate of #800, since that will address the same problem.

user12986714:
While I have no problem with this issue being closed, I still consider the displayed behaviour a bug, as going through the code base I found no reason for it to behave as such. This is likely a bug in Rails.

PR #809, opened 2020-09-09
Fix mysterious review count bug

user12986714:
I think this is the cause for #792


user12986714:
(Actually, there is no point hiding it from js if the channel provides the actual count - users can just change the js/subscribe to the channel)

PR #821, opened 2020-09-20
Revert review queue count change and remove duplicate Graph(i)QL link

double-beep:


makyen:
While it's reasonable for us to remove the GraphQL link, now that GraphiQL is working again, the links are not duplicates.

In general, please only do one thing per PR, unless they are directly connected. While we sometimes make PRs which do multiple things, it's usually better to have a PR be tightly focused and do only one thing, when that's possible.

PR #822, opened 2020-09-20
Revert "Fix mysterious review count bug"

user12986714:
Reverts Charcoal-SE/metasmoke#809


makyen:
Oh, so that was what was causing the review count to not display correctly for the last several days. I've been wondering why it was sporadically not displaying the correct number for admins. A better solution for the original "fix" this is reverting would have been to detect if the review already said "> 50+", and just not update when that was the case.

PR #823, opened 2020-09-20
Attempt to prevent duplicate feedback

user12986714:
This PR attempts to prevent duplicate feedback by adding an additional check after the database transaction is committed.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -26,6 +26,7 @@ class Feedback < ApplicationRecord
   validates :feedback_type, inclusion: { in: VALID_TYPES }
 
   after_save :populate_redis
+  after_commit :check_for_dupe_feedback_again
 
   def populate_redis
     redis.sadd "post/#{post.id}/feedbacks", id.to_s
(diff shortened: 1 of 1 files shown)


app/models/feedback.rb
```diff
@@ -182,6 +183,28 @@ def check_for_dupe_feedback
...
+                end.where(is_invalidated: false).where.not(id: id)
+    return unless duplicate.exists?
+    user = if user_id.present?
+             User.find user_id
+           else
+             User.where(user_name: user_name)
+           end
+    duplicate.each do |d|
+      if d.created_at > 1.day.ago
+        d.destroy
```
makyen, line 201: It appears you're destroying the feedback which is > 1 day old and invalidating the feedback which is < 1 day old. That's the opposite of what is desired.

user12986714: irb(main):009:0> DateTime.now > 1.day.ago
=> true

makyen: doh... brain fart on my part. Thanks.

makyen:
@ArtOfCode- Are you sure this should be merged? The failures in CI appear to be in the area directly affected by this change and, at least how I'm reading the test output, indicate that the code is not doing the correct thing in all four failing tests. It appears to be doing the same wrong thing in all four tests.

At least if I'm reading the CI test output correctly, the new code is destroying all of the user's feedback (at least in the last 24 hours) when it sees a duplicate, rather than just destroying all the feedback given by the user in the last 24 hours, *except* the most recent one (i.e. it's not leaving the most recent one).

It doesn't look like we have a test which checks for what the code does when the existing feedback is > 24 hours old, but it would probably be good to add such a test in order to double-check that the code handles that case properly and doesn't have a similar issue.

[Note: The "should not ignore unidentical feedback from the same user" test should be looking for a change of 1, not 2, but it still shows the same problem as the other two tests (i.e. that we end with 0 feedbacks from the user, when there should be 1 feedback remaining).]

user12986714:
@ArtOfCode- There is a bug introduced in this PR.

ArtOfCode-:
Cheers both - fix merged

Issue #805, opened 2020-09-04
Non-regex search results are too random


tripleee:
I don't know exactly when this changed, but the regular unrestricted MS search has basically stopped working.

It seems to do some sort of "DWIM" instead of searching for literally the phrase you put in the search box, but it never actually Does What I Mean and there doesn't seem to be any way to search for phrases with punctuation any longer.

Can we please revert back to the place we were when searching for "keto vatru" or "R-HUB" simply returned the posts which contained those phrases?

user12986714:
Theoretically you could use 'LIKE' option which is the original default behaviour.

PR #828, opened 2020-11-21
Remove fulltext index on posts

user12986714:
Reference: [Do we really need fulltext index on metasmoke?](https://stackoverflow.com/c/charcoal/questions/716)

cc #805

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -40,26 +40,10 @@ class SearchController < ApplicationController
     search_string = []
     search_params = {}
     [[:username, username, username_operation], [:title, title, title_operation],
-     [:why, why, why_operation]].each do |si|
+     [:body, body, body_operation], [:why, why, why_operation]].each do |si|
       if si[1].present? && si[1] != '%%'
         search_string << "IFNULL(`posts`.`#{si[0]}`, '') #{si[2]} :#{si[0]}"
         search_params[si[0]] = si[1]
-      end
-    end
-
-    if body.present?
-      if ['LIKE', 'NOT LIKE'].include?(body_operation) && params[:body_is_like] != '1'
-        # If the operation would be LIKE, hijack it and use our fulltext index for a search instead.
-        # UNLESS... params[:body_is_like] is set, in which case the user has explicitly specified a LIKE query.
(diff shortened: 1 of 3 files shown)


ArtOfCode-:
[status-declined]. If it's a LIKE search you need, tick the LIKE box. The fulltext index saves _vast_ amounts of time and server processing for basic searches.

user12986714:
@ArtOfCode- The issue with the fulltext index is not that it isn't blazingly fast or that it does not support CJK characters, but is that fulltext search is not doing what people (generally) mean. If I search `bbb ccc ddd` on metasmoke, I expect posts containing `bbb ccc ddd`. However, the fulltext search will give posts containing **one of** `bbb`, `ccc` or `ddd`.

Moreover, when people search `spam`, they expect posts containing string `spam`. However, in fulltext search, if a post only contains `\bspammer\b` but not `\bspam\b`, it will **not** be matched.

Consequently, the way fulltext search operates is very confusing and will be a surprise for most of the users. It operates more like Google - when we search `best spam in stackexchange` on google, we don't expect pages containing exactly `best spam in stackexchange`; google would, for example, return a page titled `best spam posted on stackexchange`.

However, on metasmoke, we have been using our old search system for years, and thus users expect their queries to return exactly posts containing exactly `best spam in stackexchange`. It is obvious from recent complaints of search inaccuracy that those users didn't seen the difference in how search operated in the past and how it operates now.

This is not about fulltext index not supporting CJK; even if all our posts are in ASCII this is still a problem.

ArtOfCode-:
@user12986714 I'm aware of the limitations of FT index searching, but the fact remains that it's saving significant processing time for the simple searches. As I said, if it's the old behaviour you need, tick the box for it - that's why it's there. Removing the index is not the correct solution.

tripleee:
Removing the full-text index may be the wrong solution, but there _is_ a problem and the result is that the behavior of search is basically erratic. Having a FAQ somewhere which explains why it's broken is by far insufficient.

Perhaps an adequate fix would be to replace the LIKE button with a "Full-text search" checkbox which is enabled by default, probably with a slightly more verbose description of how it breaks search.

user12986714:
[Edited FAQ](https://github.com/Charcoal-SE/metasmoke/wiki/FAQ:-search) to explain more on FT index.

PR #833, opened 2021-01-25
Improve search performance by removing unnecessary overheads

user12986714:
This PR tries to increase search performance (fulltext/like/regex). On my local machine, there is a 40%~60% reduction in total response time with these changes applied.

Currently, our search implementation suffers from various unnecessary overhead.
1. `will_paginate` will perform a *separate* query to count the results, which means that instead of performing a fulltext/like/regex query once, we are currently performing it *twice* for no reason.
2. There is N+1 query problem in post/_post.html.erb by using `count` method to calculate post comment counts, which will always issue a database `COUNT(*)` query regardless if the collection is already loaded.
3. We are using primitive `each` for rendering collections, while rails have a much faster collection rendering API.

---

Note: 
The `COUNT(*)` query for deciding tp/fp/naa count takes most of the time. The actual loading of posts takes only a fraction of a second.

### DIFF
--- a/app/views/posts/_post.html.erb
+++ b/app/views/posts/_post.html.erb
@@ -38,8 +38,8 @@
     <% end %>
 
     <% if post.comments.any? && !hide_feedbacks %>
-      <span class="comment-hint text-muted" title="<%= pluralize(post.comments.count, 'comment') %>">
-        <span class="glyphicon glyphicon-comment"></span> &times;<%= post.comments.count %>
+      <span class="comment-hint text-muted" title="<%= pluralize(post.comments.size, 'comment') %>">
+        <span class="glyphicon glyphicon-comment"></span> &times;<%= post.comments.size %>
       </span>
     <% end %>
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -118,7 +118,6 @@ class SearchController < ApplicationController
                   end
     end
 
-    # rubocop:disable Metrics/BlockLength
     respond_to do |format|
       format.html do
         @counts_by_accuracy_group = @results.group(:is_tp, :is_fp, :is_naa).count
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -64,7 +64,6 @@ class SearchController < ApplicationController
     end
 
     @results = @results.where(search_string.join(params[:or_search].present? ? ' OR ' : ' AND '), **search_params)
-                       .paginate(page: params[:page], per_page: per_page)
                        .order(Arel.sql('`posts`.`created_at` DESC'))
 
     @results = @results.includes(:reasons).includes(:feedbacks) if params[:option].nil?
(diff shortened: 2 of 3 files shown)


app/views/posts/_post.html.erb
```diff
@@ -38,8 +38,8 @@
     <% end %>
 
     <% if post.comments.any? && !hide_feedbacks %>
-      <span class="comment-hint text-muted" title="<%= pluralize(post.comments.count, 'comment') %>">
-        <span class="glyphicon glyphicon-comment"></span> &times;<%= post.comments.count %>
+      <span class="comment-hint text-muted" title="<%= pluralize(post.comments.size, 'comment') %>">
```
ArtOfCode-, line 41: This won't remove the additional query here. Rails overrides `size` to be the same as `count`, so it still does the query. You want to add `.includes(:comments)` to the query in the controller (or add `:comments` to an existing `includes`) , then use the normal `count` here.

ArtOfCode-, line 41: This won't remove the additional query here. Rails overrides `size` to be the same as `count`, so it still does the query. You want to add `.includes(:comments)` to the query in the controller (or add `:comments` to an existing `includes`) , then use the normal `count` here.

ArtOfCode- (changes requested):

app/views/posts/_post.html.erb
user12986714: Unless there are some versioning issues involved, AR's `Relation#size` looks like [this](https://github.com/rails/rails/blob/5f3ff60084ab5d5921ca3499814e4697f8350ee7/activerecord/lib/active_record/relation.rb#L259)
```ruby
    def size
      loaded? ? @records.length : count(:all)
    end
```
which I suppose only issues `COUNT(*)` if the records are not yet loaded?

user12986714: In addition, I believe that `:comments` are already `includes`-ed in `includes_for_post_row`

ArtOfCode-: > which I suppose only issues COUNT(*) if the records are not yet loaded?

Correct, so if it's sending a COUNT query then the records aren't loaded yet.

ArtOfCode-: Your alternative is to add a single extra query in the controller along these lines:

```ruby
@comment_counts = Comment.where(post_id: @posts.map(&:id)).group(:post_id).count
```

then in the view, instead of just `post.comments.count`:

```erb
defined?(@comment_counts) ? @comment_counts[post.id] : post.comments.count
```

user12986714: > then the records aren't loaded yet.

Pretty sure we have a misunderstanding here.

1. `Relation#count` always use a database `COUNT(*)` query regardless if the records are already loaded.
2. `Relation#size` is smarter and only uses `COUNT(*)` if the records aren't loaded yet
3. We have **already** been loading `:comments` in `Post.all.includes_for_post_row`, so changing `count` to `size` solves N+1 query problem here.

ArtOfCode-: Ah, I see where you're going. Have you checked either in Rails log or in miniprofiler that `size` does actually stop the `COUNT` queries being run?

user12986714: Those queries no longer show up in foreman logs.

PR #841, opened 2021-02-21
Always show which post in feedback messages.

makyen:
There are network delays in multiple places in the process of receiving the feedback to SD posting the message about the feedback in chat. This results in race conditions under various circumstances that, from time to time, make it unclear which post received the feedback when the message doesn't say.

In addition, now that we have a FIRE button in the feedback messages which have the post information, some people find it convenient to use that FIRE button, rather than tracking up to the associated SD report.


makyen:
Unless there are objections, I'll merge this myself. This PR exists mostly for testing.

Issue #775, opened 2020-07-16
Metasmoke is doing things multiple times


user12986714:
It appears that when performing some actions on metasmoke, it will be automatically repeated for some integer times:
- Create feedback on post: sometimes 2
- Add domain link: always 3
- Auto creation of domain: sometimes 2

This is probably not the intended behaviour of metasmoke.

Related:
https://metasmoke.erwaysoftware.com/data/sql/queries/42-duplicate-domain-tags-405
https://metasmoke.erwaysoftware.com/data/sql/queries/207-duplicate-feedbacks

ArtOfCode-:
This is probably the same old JS bug that's existed forever, probably related to Turbolinks

user12986714:
Looks like this issue is now resolved.

makyen:
The recent change I made to JavaScript event handlers may have reduced the problems, and solved some of them, but it did not resolve all of them. In particular, I can still create multiple feedbacks on the same post.

At least the multiple feedbacks, and some other issues, is the result of a user being permitted to rapidly click twice on the same control. There will be another set of JavaScript changes to resolve that, but I want to make sure the most recent change didn't result in any problems.

makyen:
The feedback issue was addressed in PR #823 and PR #825.

tripleee:
Related? #563

makyen:
Grumble, grumble, multiple things in one issue. :;

PR #848, opened 2021-03-24
Disable feedback buttons after clicking

user12986714:
Disable all feedback buttons once one of them is clicked.

There are some strange race conditions on the backend which
results in one user being able to have two or more active
(i.e. not invalidated) feedbacks on the same post, regardless
how many before_save, after_save and after_commit hooks
we run.

Ideally this should be fixed by a database trigger but
in the mean time this change prevents accidental creation
of multiple feedbacks.

Note that this change will NOT prevent intentional creation
of multiple feedbacks, as users can just call APIs directly,
circumventing restrictions implemented using javascript.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -444,3 +444,8 @@ body .profiler-results {
 .icon-nav-bar li a {
   padding: 0.7em;
 }
+
+a[data-disabled=disabled] {
+    pointer-events: none;
+    opacity: 0.5;
+}
--- a/app/javascript/feedbacks.js
+++ b/app/javascript/feedbacks.js
@@ -1,3 +1,5 @@
+import { onLoad } from './util';
+
 function disable(link) {
   link.tabIndex = -1;
   link.setAttribute('data-disabled', 'disabled');
@@ -17,20 +19,15 @@ const feedbackButtonClickHandler = () => {
   }, 2000);
 };
 
-const onLoadHandler = () => {
+onLoad(() => {
   document.querySelectorAll('.feedback-button').forEach(e => {
     e.addEventListener('click', feedbackButtonClickHandler);
   });
-};
+});
 
-const onBeforeCacheHandler = () => {
+document.addEventListener('turbolinks:before-cache', () => {
   document.querySelectorAll('.feedback-button').forEach(e => {
     e.removeEventListener('click', feedbackButtonClickHandler);
     enable(e);
   });
-  document.removeEventListener('turbolinks:load', onLoadHandler);
-  document.removeEventListener('turbolinks:before-cache', onBeforeCacheHandler);
-};
-
-document.addEventListener('turbolinks:load', onLoadHandler);
-document.addEventListener('turbolinks:before-cache', onBeforeCacheHandler);
+});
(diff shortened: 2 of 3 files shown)


makyen:
Is this really needed at this point?

I've seen nobody mention this as a problem after the most recent prior changes were made to resolve this issue. The [Duplicate feedbacks](https://metasmoke.erwaysoftware.com/data/sql/queries/207-duplicate-feedbacks) shows no results. Are there indications that this is still an ongoing issue?

makyen (changes requested):
Obviously, resolve the CI issues, at least the JavaScript ones.

Please don't use the:

    const onLoadHandler = () => {

construct, unless *required*. It is inconsistent with how most of the rest of the JavaScript is written in this project, although I do see that it's already used twice in data.js.

Don't use generic function/variable names like `clickHandler`, `onLoadHandler`, `onBeforeCacheHandler` which are in the global scope.

A helper function, `onLoad()` already exists to add `turbolinks:load` event handlers.

Feedback should not be disable on the page for infinite time. The user should be permitted to click the feedback button later, which may be to change their feedback, without having to reload the page. In other words, being able to click a feedback button should only be disabled for a short time, not forever.

I have not completely reviewed this and may have further comments.

user12986714:
For the rationale of this PR, see https://github.com/Charcoal-SE/metasmoke/issues/775#issuecomment-695606786 .
I suspect that `onLoad()` in `util.js` does not play well with turbolinks because the added listeners are not removed on `turbolinks:before-cache`.
I don't really see how generic names areparticularly problematic. If there are naming conflict concerns, that's what Webpack is for.

user12986714:
Looks like the rails and rubocop CI failure has nothing to do with this PR.

tripleee:
https://rubygems.org/gems/mimemagic/versions/0.3.5 shows this version as "yanked"; I guess the Gemfile needs to be updated in a separate PR which would block this one.

teward:
What's the logic behind this change?  Since duplicate feedback is not an issue anymore why do we need to disable the buttons after feedback?

user12986714:
@teward Duplicate feedback *is* an issue currently. That it does not present in the database is mostly because users know that they have created a duplicate feedback and deleted it. Users being cautious does not prevent duplicate feedbacks from becoming problematic however. See https://github.com/Charcoal-SE/metasmoke/issues/775#issuecomment-695606786 for more details.

makyen:
I was *strongly* under the impression that the issue with duplicate feedback was addressed by you in PRs #823 and #825. I don't recall seeing anyone complain about duplicated feedback since then. [Somewhat prior to those changes](https://chat.stackexchange.com/transcript/11540?m=55605612#55605612), I went through and invalidated all duplicate feedback, based on the ["Duplicate feedbacks" query](https://metasmoke.erwaysoftware.com/data/sql/queries/207-duplicate-feedbacks). We haven't seen any since then, as far as I'm aware.

While there are some things for which we should be de-bouncing clicks in JavaScript (thank you for reminding me that I should add and merge the code to do that generally, or at least on some *other* things), code specifically to de-bounce clicks for feedbacks doesn't appear to be needed since the changes you provided in PRs #823 and #825.

user12986714:
In this case, chances are that I failed to remenber those two PRs when looking at the issue tracker several days ago.

PR #853, opened 2021-04-11
URGENT: Upgrade mimemagic

### LINKED ISSUE #849: Dependencies update required, due to mimemagic 0.3.5 being yanked
We currently use the 0.3.5 version of the mimemagic gem. That project recently (probably 2021-03-25) yanked the 0.3.5 version from distribution. The currently available 0.3.x versions are 0.3.7 – 0.3.10. All versions have updated dependencies, so a quick drop-in manual version change in Gemfile.lock isn't sufficient. Someone with a full development environment will need to update dependencies.

user12986714:
Upgrade mimemagic to mitigate licensing concerns.

Mimemagic <= 0.3.5 contains a copy of freedesktop.org.xml, which is under GPL license.
Due to licensing concerns, those versions are now yanked from RubyGems.org.

This project is licensed under CC0, which GPL is not compatible with.

This PR upgrades Mimemagic to 0.3.10, which does not have these problems.

Package `shared-mime-info` needs to be installed on the target system. This package is already installed on most Linuxes.

closes #849

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -243,7 +243,9 @@ GEM
     mime-types (3.3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2020.0512)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
+      nokogiri (~> 1)
+      rake
     mini_mime (1.0.2)
     mini_portile2 (2.5.0)
     minitest (5.14.2)


user12986714:
There is an issue in PR #852 irrelevant to this PR.

makyen:
I'm merging this into a temporary branch to allow CI to run with the other commit from #854.

PR #854, opened 2021-04-11
Remove obsolete version of after_sign_in_path_for

user12986714:
In PR #852 there is an `after_sign_in_path_for` method added to ApplicationController. However, there is already an existing method with the same name, causing Rubocop to complain. This PR removes the old method.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,22 +29,6 @@ class ApplicationController < ActionController::Base
     raise ActionController::RoutingError, 'Not Found'
   end
 
-  def after_sign_in_path_for(resource_or_scope)
-    if current_user.username.nil?
-      users_username_path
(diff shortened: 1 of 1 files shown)


makyen:
While I generally feel that PRs should be limited to doing one thing, or a very limited number of related things, this change is really part of updating Mimemagic. Having it as a separate PR that is based on the current master branch makes it so we can't actually see if *either* of these PRs actually fully fixes the problems which they set out to fix (i.e. we will only see CI actually pass or fail once both of them are merged).

Given that this is directly necessary due to the dependency updates which go along with updating Mimemagic, the commit included here should be part of PR #853. Having it there should allow CI to fully run on the combination of the two commits and would tell us if the changes actually fix the problems (or, at least the problems with CI). That way we will know if anything else is needed.

makyen:
I'm merging this into a temporary branch to allow CI to run with the other commit from #853.

PR #857, opened 2021-04-15
Add abusive flag

### LINKED ISSUE #722: Add a "Rude flag" link next to the "Spam flag" link
image
At the very end of a MS report, there is a link of which one can cast a spam flag. Would we consider adding a "Rude flag" link in addition to that, as it is suggested not to use spam flags for rude/abusive content (and a user may not want to firstly reply k in chat and then go to the site to flag)?

user12986714:
closes #722

### DIFF
--- a/app/assets/stylesheets/posts.scss
+++ b/app/assets/stylesheets/posts.scss
@@ -35,3 +35,8 @@ body.posts .bootstrap-select .btn.dropdown-toggle {
   padding-top: 4px;
   padding-bottom: 4px;
 }
+
+.flag-link {
+  margin-left: 2em;
+  margin-right: 2em;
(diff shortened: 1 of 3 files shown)


Undo1:
This looks good to me at a glance, will let someone more familiar with the codebase also review. 

As a sidenote, I noticed we have fewer strings for spam flags (4) than the other types (5 each). Maybe something we should investigate? https://github.com/Charcoal-SE/metasmoke/blob/master/app/models/user.rb#L200

makyen:
@Undo1 4 different strings is correct for spam flags, whereas the other types of flags have 5. Both English & Portuguese use "spam" for the string.

Issue #900, opened 2022-02-02
database is too big for Docker


tripleee:
When importing the latest database, I bumped into

```
 => [14/21] COPY . .                                                       8.2s
 => [15/21] RUN gem install bundler                                        2.0s
 => [16/21] RUN bundle update --bundler                                  211.1s
 => [17/21] RUN bundle install                                             2.0s
 => ERROR [18/21] RUN ./createdb                                           2.0s
------
 > [18/21] RUN ./createdb:
#22 0.440 ./createdb: importing import/redis_1643761044.rdb.gz
#22 1.867 /etc/init.d/mariadb: ERROR: The partition with /var/lib/mysql is too full! ... failed!
------
executor failed running [/bin/sh -c ./createdb]: exit code: 1
```

There seems to be a workaround of sorts in https://stackoverflow.com/questions/37100065/resize-disk-usage-of-a-docker-container but it apparently affects all your Docker containers.

tripleee:
Notice that the `Dockerfile` does `COPY . .` so if you have temporary files you are not using for anything (such as previous imports!) those will be copied into the image, too.

teward:
I'm confused, are you saying that this is a problem with the database being large, or that it's a problem with the Docker image?

If this is an issue with Docker containers not being provisioned large enough, that sounds more like an issue with your Docker daemon not provisioning enough and thus something that has to be handled on the Docker daemon side of things.  just my 2 cents.

tripleee:
> If this is an issue with Docker containers not being provisioned large enough, that sounds more like an issue with your Docker daemon not provisioning enough and thus something that has to be handled on the Docker daemon side of things. just my 2 cents.

This is indeed the obvious workaround, but Docker out of the box defaults to 10G disk images; it would be nice to be able to stay within this limit.

PR #901, opened 2022-02-03
createdb: remove import files to save disk space

tripleee:
(closes: #900) or at least alleviates the problem somewhat

### DIFF
--- a/createdb
+++ b/createdb
@@ -9,6 +9,7 @@ if [ -d import ]; then
             gzip -dc "$redis" >dump.rdb
             chmod 644 dump.rdb
             # chown redis:redis dump.rdb
+            rm -f "$redis"
         else
             echo "$0: No files matching $redis" >&2
             exit 124
@@ -66,6 +67,7 @@ if [ -d import ]; then
             gzip -dc <"$sql" |
             sed 's/dump_metasmoke/metasmoke/g' |
             mysql -u root -ppassword metasmoke
+            rm -f "$sql"
         else
             echo "$0: No files matching $sql" >&2
             exit 123
--- a/createdb
+++ b/createdb
@@ -2,6 +2,12 @@
 
 set -e -u
 
+case $(pwd) in
+ /usr/src/app) ;;
+ *) echo "$0: this should only be run from the Dockerfile inside the image" >&2
+    exit 33;;
+esac
+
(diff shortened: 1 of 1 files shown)


makyen:
This appears to be deleting the original database backup files. I don't like the idea of auto-deleting these files which the build process didn't create, particularly doing so without notice to the human and without the human being given the option not to do so. If these were my only copies of the database backup files, I'd be fairly pissed that they were auto-deleted. I might be using a database backup from a specific prior date, which could be impossible to replace.

makyen:
The parent commit for these changes is 3 months old. Please pull the current version of the repository's master branch and rebase your changes to that commit, so that CI testing functions correctly.

If you choose not to do that, then given how GitHub performs merges, please don't use "create a merge commit". Please select "Rebase and merge" when choosing how to merge. GitHub generally does a merge of the master branch into the PR, rather than just the PR into master. Doing that can result in a commit which is confusing when looking at the diffs, as it will look like you're re-making every change between the commit upon which this is parented and the current HEAD of the master branch.

tripleee:
> This appears to be deleting the original database backup files. I don't like the idea of auto-deleting these files which the build process didn't create, particularly doing so without notice to the human and without the human being given the option not to do so. If these were my only copies of the database backup files, I'd be fairly pissed that they were auto-deleted. I might be using a database backup from a specific prior date, which could be impossible to replace.

The script supposed to run inside the Docker image, which contains a copy from the `COPY . .` earlier in the file.  Perhaps there should be a guard against running it outside the image, though.

tripleee:
> The parent commit for these changes is 3 months old. Please pull the current version of the repository's master branch and rebase your changes to that commit, so that CI testing functions correctly.

Done, thanks. Sorry, I thought I pulled upstream before creating a branch, but apparently not.

makyen:
It doesn't appear that the current version of the repository was pulled and `git rebase master` run. The [parent commit](https://github.com/Charcoal-SE/metasmoke/commit/7bb8431e3e638cbbed06e75c7b54b86f52d48ecc) for this PR is still the same.

tripleee:
Grrr, I had used really dumb remote names in the distant past which tripped me up. That also explains why the branch for this got pushed to the main repo, rather than my personal fork.

PR #904, opened 2022-02-06
Spam domain API fix

tripleee:
As suggested by @user12986714, parse out the argument from `request.env['PATH_INFO']` in the `name/:name` for `spam_domains` and posts `uid/:api_param/:native_id` for `posts` routes.

### DIFF
--- a/app/api/api/posts_api.rb
+++ b/app/api/api/posts_api.rb
@@ -83,10 +83,13 @@ module API
       std_result Post.where(deleted_at: nil).order(id: :desc), filter: FILTERS[:posts]
     end
 
-    # We are not using "get 'uid/:api_param/:native_id' do", because that does not permit a literal '.' in the api param,
-    # which is required for meta and localized SO sites.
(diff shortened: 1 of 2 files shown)


tripleee:
I wanted to create a test case for this regression but it seems there are no test cases for the API and since this is my first attempt at writing Ruby, it felt too overwhelming. )-:

tripleee:
Argh, concessions to the unforgiving Rubocop.

tripleee:
I also looked at https://stackoverflow.com/questions/2952235/rails-params-with-dot-e-g-google-com but was unable to do anything useful with the answers there. In particular,

```
    get 'name/:name', constraints: { name: /^\w+$/ } do
```
did not produce anything useful. I also tried to play around with the idea to add optional parameters from https://stackoverflow.com/questions/47758775/how-to-apply-query-parameter-constraints-to-grape but I probably botched it somehow, or maybe that idea doesn't work inside an API handler of our type either.

PR #952, opened 2022-10-24
Return search & flagging to topbar while still being responsive and not expanding the topbar vertically.

makyen:
The search and flagging navigation buttons in the icon bar at the top-left obscure page content at some viewport widths. In addition, at least to me, having them in the left icon bar has always felt a bit clunky and doesn't feel like it really "fits" with the rest of the site design.

This PR returns the search and flagging selections to the topbar from the icon bar on the top-left. The stated reason for moving them to an icon bar was so the topbar didn't wrap (and take up substantial additional vertical space). This makes the topbar  responsive through the widths which are not put into a hamburger menu (i.e, viewports >=768px wide), such that the buttons in the topbar don't wrap. However, if the user's MS username is too long it will wrap into a second line, which is kept within the existing height of the topbar. If the user's MS username is too long to fit into two lines, then it's truncated with an ellipse.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -448,3 +448,140 @@ body .profiler-results {
 .wait-for-AJAX.wait-for-AJAX {
   cursor: progress;
 }
+
+/* Search and flagging back in topbar, but responsive */
+.navbar-static-top {
+  .navbar-nav > li > a.glyphicon {
+      display: none;
+  }
+  @media (min-width: 768px) {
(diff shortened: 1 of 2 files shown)


makyen:
@thesecretmaster Just FYI: This is just a heads-up. You're the one who [moved the search and flagging selections out of the topbar](https://github.com/Charcoal-SE/metasmoke/commit/25d65543fd9b528c7010aecbde3b925363c498a0). I believe this solves the issue you were trying to address, while keeping those two nav selections in topbar.

PR #976, opened 2023-03-31
Add a regex_cache to the SpamWave model and end post_matches? early

makyen:
The primary issue this PR is attempting to improve is to reduce the time it takes to validate a spam waves, particularly the support number spam wave. The Support Number spam wave takes a considerable amount of time to validate (many tens of minutes) with a reduced complexity regular expression, and will often fail to validate due to taking too long. This PR should also reduce the CPU load for testing each reported post against active spam waves.

For regular expressions, a considerable amount of the time to use one is the time it takes to compile it into the representation which the regular expression engine then uses to run the regex. Currently, the spam wave code compiles each of the three regular expressions in the spam wave every time a post is run through `post_matches?()`. For validation, that can be several thousands of post reports. That's quite a bit of extra time spent re-compiling the same regexes.

This PR adds two levels of caching to the regexes used in the the `SpamWave` class. The first level is `regex_cache` as a class variable. If there's no compiled `Regexp` in the `regex_cache` for the regular expression text, then `Rails.cache` is checked for a serialized version of the compiled `Regexp`. If it doesn't exist in `Rails.cache`, then the regex text is compiled and the Regexp is stored in both levels of cache. The size of `regex_cache` is limited to `MAX_REGEX_CACHE_SIZE`, which is, currently, set to 100. Entries in the `Rails.cache` expire after 6 hours.

In addition to immediately returning false upon failure to match the non-regex based criteria, `false` is returned at the first point one of the three regexes fails to match. The regexes are run against the post in the order of increasing maximum length of the text (i.e. username, title, body) in order to fail with the least amount of effort, if possible.

The overall expectation is a notable reduction in the amount of time it takes to validate a spam wave and reduce the amount of time it takes to determine if an incoming report matches each active spam wave.

### DIFF
--- a/app/models/spam_wave.rb
+++ b/app/models/spam_wave.rb
@@ -14,6 +14,16 @@ class SpamWave < ApplicationRecord
   scope(:active, -> { where('expiry >= ?', DateTime.now) })
   scope(:expired, -> { where('expiry < ?', DateTime.now) })
 
+  # The regex_cache won't be consistent across threads, but we don't need it
+  # to be. We only need it to not be corrupted from uses across threads, which it
+  # shouldn't be, as threads don't share class variables.
+  #
+  # In addition, a Rails.cache entry is created for each Regexp. That entry should
+  # be available across threads, but will be a bit slower due to the
+  # serialization/deserialization steps.
+  @@regex_cache = {} # rubocop:disable Style/ClassVars
+  MAX_REGEX_CACHE_SIZE = 100
+
   def post_matches?(post, site_ids = nil)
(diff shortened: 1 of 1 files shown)


makyen:
While I'm interested in having this reviewed and generally prefer not to merge without giving at least 24 hours for review, I'm going to go ahead and merge this and deploy, then watch the logs to verify it's doing what it's supposed to do. Given the current activity with support number spam, it would be quite helpful to have the support number spam wave working, which this will hopefully allow. Around now is also the lowest activity day/time of the week, so hopefully the least impact, should there be problems.

Issue #924, opened 2022-05-19
Clarify user documentation around spam wave use of autoflagging permissions


tripleee:
Just reiterating a discussion from Charcoal.

https://stackoverflow.com/c/charcoal/questions/868

In brief, update the documentation strings in metasmoke to clarify that opting in to autoflagging also means your flags will be eligible for use in spam waves, irrespective of your threshold settings etc for regular autoflagging.

gparyani:
I highly agree with documenting this, given that some users have [raised complaints](https://stackoverflow.com/c/charcoal/questions/632) about their flags being used in spam waves without anything pointing them to that fact.

PR #981, opened 2023-05-27
Add a note mentioning spam waves to the Autoflagging Preferences page

CoconutMacaroon:
## Rationale

Autoflags _for spam waves specifically_ are automatically cast for users with the Checkbox to enable MS flagging, regardless of their autoflagging preferences (or lack thereof). This should be documented somewhere, as it is bad practice to cast flags without giving them a heads-up first, especially as it isn't explained anywhere.

Spam wave autoflagging has also been discussed multiple times in CHQ.

## TODO
- [x] Write text explaining the existence of spam wave autoflagging
- [x] Add a link to more details on what spam waves are
- [ ] Test these changes (I haven't yet)

### DIFF
--- a/app/views/user_site_settings/index.html.erb
+++ b/app/views/user_site_settings/index.html.erb
@@ -33,7 +33,7 @@
       A) this checkbox must be checked;
       B) you must have at least one valid autoflagging preference (see "Preferences" below); <em>and</em>
       C) you must have at least one valid <%= link_to "autoflagging condition", url_for(controller: :flag_conditions, action: :index) %>.
-      Importantly, by checking the above box, autoflags may be automatically cast from your account for spam waves, regardless of your autoflagging settings (if any). Further note that there is not currently a way to opt-out of this. If the above checkbox is checked and you have either or both no valid autoflagging preferences or no valid and enabled  autoflagging conditions,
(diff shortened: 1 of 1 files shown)


CoconutMacaroon:
Once completed, this should fix #924

Spevacus:
I'd adjust the full blurb a bit more. Specifically, I'd change it to:

> If the above checkbox is checked and you have either or both no valid autoflagging preferences or no valid and enabled autoflagging conditions, then autoflagging for your account will be disabled and only manual flagging will be functional.
> 
> There is one exception to this. By checking the above box, autoflags may be automatically cast from your account for spam waves, regardless of your autoflagging settings. Spam waves are a manual tool used sparingly by Metasmoke admins to fight waves of spam or abusive content, and can be configured to cast the necessary number of spam flags to instantly delete posts that meet specific criteria.

CoconutMacaroon:
I tried to test these changes. Unfortunately, metasmoke wants me to authenticate with SE before showing that text, and I had trouble doing that. As a result, I have not successfully tested this.

makyen (changes requested):
I don't like this with the following text:
>, and can be configured to cast the necessary number of spam flags to instantly delete posts that meet specific criteria.

While it's *possible* for a spam wave to be configured to raise enough autoflags to delete a post, that's *very rarely* done and only done with the express permission from either site moderators on the specific site that's affected or from CMs when it affects the entire network. The way it's phrased here implies that we are free to completely auto-nuke whatever the admins feel like. If the required approvals are not going to be mentioned, then the entire phrase shouldn't be stated.

talflon:
I like this being documented so would like to see this merged. How about something like:

> and in extreme cases can be configured (when expressly authorized by site mods or CMs) to cast ...

Issue #934, opened 2022-08-16
Time to deletion chart seems broken


tripleee:
Is it just me? https://metasmoke.erwaysoftware.com/sites/dash?utf8=%E2%9C%93&site_id=21&months=3&tab=all displays "Error Loading Chart: Internal Server Error" for the TTD chart.

<img width="1141" alt="Screenshot 2022-08-16 at 13 03 54" src="https://user-images.githubusercontent.com/2160915/184853517-7f8b9b20-95c1-4137-a691-de03fac0c177.png">

makyen:
For reference, the error text returned from the GET to `https://metasmoke.erwaysoftware.com/graphs/monthly_ttd?months=3&round=0&site_id=21` is

```
NoMethodError in GraphsController#monthly_ttd

undefined method `round' for nil:NilClass
Extracted source (around line #161):

#159              end
#160            end
*161     data = data.map { |k, v| [k, v.round(params[:round].to_i)] } if params[:round]
#162     render json: data
#163   end
#164 

Extracted source (around line #161):

#159              end
#160            end
*161     data = data.map { |k, v| [k, v.round(params[:round].to_i)] } if params[:round]
#162     render json: data
#163   end
#164 

Extracted source (around line #161):

#159              end
#160            end
*161     data = data.map { |k, v| [k, v.round(params[:round].to_i)] } if params[:round]
#162     render json: data
#163   end
#164 


Rails.root: /var/railsapps/metasmoke/releases/20220814172854

Application Trace
app/controllers/graphs_controller.rb:161:in `block in monthly_ttd'
app/controllers/graphs_controller.rb:161:in `each'
app/controllers/graphs_controller.rb:161:in `map'
app/controllers/graphs_controller.rb:161:in `monthly_ttd'

Framework Trace
actionpack (5.2.6.2) lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
actionpack (5.2.6.2) lib/abstract_controller/base.rb:194:in `process_action'
actionpack (5.2.6.2) lib/action_controller/metal/rendering.rb:30:in `process_action'
actionpack (5.2.6.2) lib/abstract_controller/callbacks.rb:42:in `block in process_action'
activesupport (5.2.6.2) lib/active_support/callbacks.rb:109:in `block in run_callbacks'
audited (4.10.0) lib/audited/sweeper.rb:14:in `around'
[...]

PR #1025, opened 2024-11-20
Bypass nil values when calculating TTD chart

tripleee:
Closes: #934


tripleee:
Literally a single-character change, based on https://stackoverflow.com/a/36631581/874188

